PR #15, opened 2016-07-14
Adding menutype when displaying menu items (Patch from JM)

jreys:
#### Summary of Changes

Adding menutype when displaying menu items
#### Testing Instructions

Navigate administrator/index.php?option=com_associations&view=associations and select menu items and one of the languages, there's now a column that shows menu types

### DIFF
--- a/administrator/components/com_associations/models/associations.php
+++ b/administrator/components/com_associations/models/associations.php
@@ -33,6 +33,7 @@ class AssociationsModelAssociations extends JModelList
 				'title',
 				'ordering',
 				'level', 'a.level',
+				'menutype', 'a.menutype',
 				'association',
 				'associationlanguage',
 				'associationcomponent',
@@ -110,6 +111,7 @@ class AssociationsModelAssociations extends JModelList
 		$component      = $this->getState('associationcomponent');
 		$table          = '';
 		$extension      = '';
+		$menuType       = false;
 
 		// If it's not a category
 		if (!strpos($component, '|'))
@@ -128,6 +130,7 @@ class AssociationsModelAssociations extends JModelList
 				if ($componentSplit[0] == 'com_menus')
 				{
 					$table = '#__menu';
+					$menuType = true;
 				}
 				else
 				{
@@ -154,6 +157,11 @@ class AssociationsModelAssociations extends JModelList
 		if (isset($columns['level']))
 		{
 			$query->select($db->quoteName('a.level'));
+		}
+
+		if ($table == '#__menu')
+		{
+			$query->select($db->quoteName('a.menutype'));
 		}
 
 		$query->from($db->quoteName($table, 'a'));
@@ -224,6 +232,11 @@ class AssociationsModelAssociations extends JModelList
 		}
 
 		// Add the list ordering clause.
+		if ($table != '#__menu')
+		{
+			$this->setState('list.ordering', 'ordering');
+		}
+
 		$orderCol = $this->state->get('list.ordering', 'ordering');
 		$orderDirn = $this->state->get('list.direction', 'asc');
 
--- a/administrator/components/com_associations/models/associations.php
+++ b/administrator/components/com_associations/models/associations.php
@@ -111,7 +111,6 @@ class AssociationsModelAssociations extends JModelList
 		$component      = $this->getState('associationcomponent');
 		$table          = '';
 		$extension      = '';
-		$menuType       = false;
 
 		// If it's not a category
 		if (!strpos($component, '|'))
@@ -130,7 +129,6 @@ class AssociationsModelAssociations extends JModelList
 				if ($componentSplit[0] == 'com_menus')
 				{
 					$table = '#__menu';
-					$menuType = true;
 				}
 				else
 				{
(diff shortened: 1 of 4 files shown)


administrator/components/com_associations/models/associations.php
```diff
@@ -224,6 +232,11 @@ protected function getListQuery()
 		}
 
 		// Add the list ordering clause.
+		if ($table != '#__menu')
+		{
+			$this->setState('list.ordering', 'ordering');
+		}
```
andrepereiradasilva, line 236: ?

administrator/components/com_associations/views/associations/tmpl/default.php
```diff
@@ -16,6 +16,13 @@
 
 $listOrder = $this->escape($this->state->get('list.ordering'));
 $listDirn  = $this->escape($this->state->get('list.direction'));
+
+$colSpan = $this->menuType == true ? 5 : 4;
+
+if ($this->menuType == '')
+{
+	$colSpan--;
+}
```
andrepereiradasilva: ?

jreys: colspan should be dynamic, 4 when you don't have the menutype column, 5 when you have it

andrepereiradasilva: true, so you just need

``` php
$colSpan = $this->menuType === true ? 5 : 4;
```

right?

administrator/components/com_associations/models/associations.php
jreys, line 236: @infograf768 can you explain please? :sweat_smile:

administrator/components/com_associations/views/associations/tmpl/default.php
```diff
@@ -53,7 +65,7 @@
 			</thead>
 			<tfoot>
 				<tr>
-					<td colspan="4">
+					<td colspan="<?php echo $colSpan; ?>">
```
jreys, line 63: yes, @andrepereiradasilva check here

administrator/components/com_associations/models/forms/filter_associations.xml
```diff
@@ -52,6 +51,8 @@
 				>
 				JASSOCIATIONS_DESC
 			</option>
+			<option value="a.menutype ASC">COM_MENUS_HEADING_MENU_ASC</option>
+			<option value="a.menutype DESC">COM_MENUS_HEADING_MENU_DESC</option>
```
andrepereiradasilva, line 55: hum maybe we need something like `requires` for disabling options is the list options...
IMO this menu ordering should only be available when menu component is selected.

I'm thinking of `showon`, but i think it doesn't work in option tag

administrator/components/com_associations/views/associations/tmpl/default.php
andrepereiradasilva, line 63: joao please read the code!

``` php
$colSpan = $this->menuType == true ? 5 : 4;
```

so, colspan is 5 or 4 depending if we seleted menu item. ok correct

But ...

``` php
if ($this->menuType == '')
{
    $colSpan--;
}
```

... if this var is empty we remove yet another column from the colspan? why?

administrator/components/com_associations/models/associations.php
infograf768, line 236: The problem is the following:
If we display menu items and sort by menu ordering and then switch to another item than menu items (articles, categories), we get an sql error. This is the only way I found to avoid this issue. Better way welcome.

administrator/components/com_associations/views/associations/tmpl/default.php
infograf768, line 63: Indeed, may be useless. Was just a copy/paste

administrator/components/com_associations/models/forms/filter_associations.xml
infograf768, line 55: See above, as this is the reason of the setState.

administrator/components/com_associations/models/associations.php
```diff
@@ -110,6 +111,7 @@ protected function getListQuery()
 		$component      = $this->getState('associationcomponent');
 		$table          = '';
 		$extension      = '';
+		$menuType       = false;
```
infograf768: Can be taken off. Remnants of a different way to select menutype

administrator/components/com_associations/models/associations.php
```diff
@@ -128,6 +130,7 @@ protected function getListQuery()
 				if ($componentSplit[0] == 'com_menus')
 				{
 					$table = '#__menu';
+					$menuType = true;
```
infograf768: Can be taken off. Remnants of a different way to select menutype

administrator/components/com_associations/models/forms/filter_associations.xml
andrepereiradasilva, line 55: Maybe we can add this (and others if needed) options dinamicly as we do for the column ?

administrator/components/com_associations/models/associations.php
andrepereiradasilva, line 236: ok we need to check that

andrepereiradasilva:
i will merge this and we will improve later

PR #16, opened 2016-07-14
Moving CSS from view to template

jreys:
#### Summary of Changes

Moving CSS from com_associations/views/association/view.html.php to com_associations/views/association/tmpl/edit.php

### DIFF
--- a/administrator/components/com_associations/views/association/view.html.php
+++ b/administrator/components/com_associations/views/association/view.html.php
@@ -92,44 +92,9 @@ class AssociationsViewAssociation extends JViewLegacy
 			}
 		}
 
-		$app = JFactory::getApplication();
-		$input = $app->input;
+		$this->app = JFactory::getApplication();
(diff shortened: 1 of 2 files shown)


administrator/components/com_associations/views/association/view.html.php
```diff
@@ -92,45 +92,10 @@ public function display($tpl = null)
 			}
 		}
 
-		$app = JFactory::getApplication();
-		$input = $app->input;
+		$this->app = JFactory::getApplication();
+		$input = $this->app->input;
```
andrepereiradasilva, line 96: $this->app ? humm

jreys, line 96: Using the same var on the view and template

andrepereiradasilva, line 96: ok not used to it. but i see no problem on that

Issue #7, opened 2016-06-07
ACL across com_associations


andrepereiradasilva:
#### Introduction

This issue is a placeholder for ACL implementation across the com_associations component.
This issues serves as brainstorm and project planing.

We will needs several types of ACL check:
##### com_associations ACL

com_associations compoent ACL should be used only for the component itself.
- "Configure ACL & Options" (core.admin)
- "Configure Options Only" (core.options)
- "Access Administration Interface" (core.manage)

This ACL options should work exactly like other components.

Task level: Easy

**[Already done!]**
##### com_languages language item ACL **[decided to not go this way]**

This is still in brainstrom, but probaly there will exist a "Use" (core.use) ACL in each language that will disable selecting that language in the language select boxes.

So probably, since the select box are fields generated by the joomla api, this component will not need to do anything.

Task level: Probably none
##### com_`<component>` `<item>` ACL

List and side by side view must respect the com_`<component>` `<item>` ACL

Some examples:
- If a user does not have access to "Access Administration Interface" (core.manage) com_contacts, that component can't appear in the components list of com_associations.
- If a user does not have access to "Edit" (core.edit) com_contacts, the user can't edit the associations of the contacts, but can view them and create new associations.
- If a user does not have access to "Create" (core.create) com_contacts, the user can't create new associations for a particular contact, but can view them and edit them.

Task level: Medium

infograf768:
> This is still in brainstrom, but probaly there will exist a "Use" (core.use) ACL in each language that will disable selecting that language in the language select boxes.

I chose finally `core.permission` and a project PR is here (without the new lang strings)
Look here:
https://github.com/infograf768/joomla-cms/pull/30

andrepereiradasilva:
where we are on this?

andrepereiradasilva:
hi all, basic ACL in the list view for me is the easy part. we already have the item ACL key in the helper 
https://github.com/joomla-projects/gsoc16_improved-multi-lingual/blob/staging/administrator/components/com_associations/helpers/associations.php#L135

you just need to call the helper
example for categories in the list helper

``` php
$component = AssociationsHelper::getComponentProperties('com_categories.category|com_content');
$user      = JFactory::getUser();
$userId    = $user->get('id');
// [...]
<?php foreach ($this->items as $i => $item) : ?>
    <?php
    $canCreate  = $user->authorise('core.create',     $component->assetKey . $item->id);
[...]

Issue #1, opened 2016-05-07
Prepare Article for JCM


puneet0191:
Get it reviewed from Mentors, Coordinate with Ana and get the article into the submission bucket.

jreys:
Done Puneet, you can close :)

puneet0191:
Thanks Completed :+1:

PR #17, opened 2016-07-14
Complete side-by-side view and implemented language switch

jreys:
#### Summary of Changes

Complete side-by-side view and implemented language switch


jreys:
It seems the "save" option is only working in contacts by using the JS inside.getElementById('applyBtn').click(); on other components it's not working

andrepereiradasilva:
fix conlicts please

andrepereiradasilva:
IMO the save button need to be in the component toolbar like all other componets. UX

jreys:
yes, it's just for testing purposes, it will be replaced by the one on the top

andrepereiradasilva:
ok, then fix the conflicts please

jreys:
Fixed conflicts

administrator/components/com_associations/views/association/view.html.php
```diff
@@ -92,32 +92,51 @@ public function display($tpl = null)
...
+=======
+		$this->form  = $this->get('Form');
+
+		$this->app = JFactory::getApplication();
+		$input     = $this->app->input;
+		$assoc     = JLanguageAssociations::isEnabled();
+		
+		$formData             = new JInput($input->get('jform', '', 'array'));
+		$targetId = $formData->getInt('itemlanguage', 0);
+>>>>>>> side-by-side
```
andrepereiradasilva: does it seem fixed to you @jreys ?

jreys: now it is, sorry

infograf768:
I can't apply diff for views/association/view.html.php

Also I don't really understand why we have `&layout=modal` in the side by side.
Could you explain what you want to do?
Why do we still have Apply/save/etc. in the toolbar in the side by side view? I really think Apply and Save should be only on Top of the Target items.

andrepereiradasilva:
@jreys reggaring the modal see https://github.com/jreys/gsoc16_improved-multi-lingual/pull/4/files. you cannot merge it because it conflicts, but you can use that code for a commit and close the PR.

andrepereiradasilva:
> Why do we still have Apply/save/etc. in the toolbar in the side by side view? I really think Apply and Save should be only on Top of the Target items.

@infograf768 IMHO from a UX perspective the buttons should be where they are in all joomla backend views, ie, in the toolbar.

infograf768:
> @infograf768 IMHO from a UX perspective the buttons should be where they are in all joomla backend views, ie, in the toolbar.

I respectfully disagree. Showing it on top of what is going to be applied or saved makes much more sense. It is a very special UX here (side by side) where we do not switch to another page after saving but will most likely through the language dropdown display an existing or possible association with another item in another language.

andrepereiradasilva:
[...]
1. BTW i see some future improvements for this component
   - Add a modal layout also to this view that could be used it in "Associations" tab "New"/"Edit" button of the edit item in the components, in this case the Save/Close/Etc butons, like all other modals, would be on modal bottom

andrepereiradasilva:
joao please fix cs and implement the other changes you say it's to do in description

administrator/components/com_associations/models/fields/itemlanguage.php
```diff
@@ -0,0 +1,92 @@
...
+				$associations = call_user_func(array($helpAssoc, 'getAssociations'), $referenceId, $associatedView);
+				$existingLanguages = JHtml::_('contentlanguage.existing', false, true);
+
+				foreach ($existingLanguages as $key => $lang)
+				{
+					if ($lang->value == $forcedLanguage)
+					{
+						unset($existingLanguages[$key]);
+					}
+ 					if (isset($associations[$lang->value]))
```
infograf768: What is that part of the code for?

administrator/components/com_associations/models/fields/itemlanguage.php
```diff
@@ -0,0 +1,92 @@
...
+		$options = array();
+
+		$input = JFactory::getApplication()->input;
+
+		$referenceId         = $input->get('id', '');
+		$associatedComponent = $input->get('acomponent', '');
+		$associatedView      = $input->get('aview', '');
+		$forcedLanguage      = $input->get('forcedlanguage', '');
+
+		if ($associatedComponent == 'com_categories')
```
infograf768: I used here for categories something like:

```
if (file_exists(JPATH_SITE . '/components/' . $associatedView . '/helpers/association.php'))
            {
                $componentName = ucfirst(substr($associatedView, 4));
                JLoader::register($componentName . 'HelperAssociation', JPATH_SITE . '/components/' . $associatedView . '/helpers/association.php');

                if (method_exists($componentName . 'HelperAssociation', 'getCategoryAssociations'))
                {
                    $associations = call_user_func(array($componentName . 'HelperAssociation', 'getCategoryAssociations'), $referenceId, $associatedView);
                    $existingLanguages = JHtml::_('contentlanguage.existing', false, true);

                    foreach ($existingLanguages as $key => $lang)
                    {
                        if ($lang->value == $forcedLanguage)
                        {
                            unset($existingLanguages[$key]);
                        }
[...] =>don't know what the rest is for, see below.
```

administrator/components/com_associations/models/fields/itemlanguage.php
```diff
@@ -0,0 +1,92 @@
...
+		{
+
+		}
+		elseif ($associatedComponent == 'com_menus')
```
infograf768: I used for com_menus something like:

```
$helpAssoc = 'MenusHelper';
            JLoader::register($helpAssoc, JPATH_ADMINISTRATOR . '/components/com_menus/helpers/menus.php');
            $associations = call_user_func(array($helpAssoc, 'getAssociations'), $referenceId);
            $existingLanguages = JHtml::_('contentlanguage.existing', false, true);

                foreach ($existingLanguages as $key => $lang)
                {
                    if ($lang->value == $forcedLanguage)
                    {
                        unset($existingLanguages[$key]);
                    }
[...]
```

administrator/components/com_associations/models/fields/itemlanguage.php
jreys: This part fetches each item's associations, it loads the helper located on the front end (getAssociations) because the global method needs the item's table. if you're talking about if ($lang->value == $forcedLanguage), it's to remove the forced language from the language switcher since it's not needed.

P.S. - The router is being loaded because it is called on the getAssociations function and without loading it, it throws out an exception.

infograf768: I was speaking of

```
if (isset($associations[$lang->value]))
{
+                       parse_str($associations[$lang->value], $contents);
+                       $associatedID = $contents['id'];
+                       $removeExtra = explode(":", $associatedID);
+                       $lang->value = $removeExtra[0];
+                   }
+                   else
+                   {
+                       $lang->value = 0;
+                   }
```

jreys: I did this to extract the item's ID, it seems when I called that method, besides de id, it came with the alias too, id:alias, etc., I just need the ID in this case, if it's not present I'll just use the ID 0, which opens an empty editable item, of course this part needs some tweaks.

infograf768: Not sure this works as intended as I did not use it for menus and categories and I got empty stuff when no assoc

jreys: That's the magic part :smiley: , you want a new item when there's no association and when it's saved it will associate to the reference in that language, am I right? currently that behaviour is not implemented, but it's a start.

administrator/components/com_associations/views/association/tmpl/edit.php
```diff
@@ -67,7 +67,7 @@ function triggerSave() {
 ?>
 <button id="toogle-left-panel">Show/Hide Reference (PoC)</button>
 
-<form action="<?php echo JRoute::_('index.php?option=com_associations&view=association'); ?>"
+<form action="<?php echo JRoute::_(JFactory::getURI()->toString()); ?>"
```
andrepereiradasilva: it's better to add the all URL parameters inside JRoute, like all other views.
We need always to be consistent with the rest of the code in the core.

administrator/components/com_associations/views/association/view.html.php
```diff
@@ -93,32 +93,36 @@ public function display($tpl = null)
...
-		if ($associatedComponent == 'com_categories')
-		{
-			// If it's categories
-			$this->link = 'index.php?option=' . $associatedComponent . '&task=category.edit&layout=modal&tmpl=component&id='
-				. $referenceId . '&extension=' . $associatedView;
-		}
-		elseif ($associatedComponent == 'com_menus')
+		$this->form  = $this->get('Form');
+		$input     = $this->app->input;
+		$assoc     = JLanguageAssociations::isEnabled();
```
andrepereiradasilva: why do we need this here?

administrator/components/com_associations/models/fields/itemlanguage.php
jreys: Edited it a little but as it is I don't know how you found it working :stuck_out_tongue_closed_eyes:

administrator/components/com_associations/views/association/tmpl/edit.php
```diff
@@ -22,7 +22,7 @@ function iframeRef( frameRef ) {
 
 	function triggerSave() {
 		var inside = iframeRef( document.getElementById('target-association') );
-		inside.getElementById('applyBtn').click();
+		window.frames[1].Joomla.submitbutton('article.apply');
```
jreys: The save action is working but it doesn't load the layout to modal again, my previous implementation with the "inside.getElementById('applyBtn').click();" worked only for contacts, my questions are:
- Can my current implementation be tweaked?
- Should other core components implement my older implementation?
- Are both of these solutions wrong?

administrator/components/com_associations/views/association/tmpl/edit.php
```diff
@@ -82,16 +82,15 @@ function triggerSave() {
 	<div class="outer-panel" id="right-panel">
 		<div class="inner-panel">
 			<h3><?php echo JText::_('COM_ASSOCIATIONS_ASSOCIATED_ITEM'); ?></h3>
-			<select id="ref-language" name="ref-language">
-				<?php echo JHtml::_('select.options', JHtml::_('contentlanguage.existing', false, true), 'value', 'text'); ?>
-			</select>
+			<?php echo $this->form->getInput('itemlanguage'); ?>
```
jreys: How do I save this input's state so when I reload the page it's pre-selected with the loaded language?

infograf768:
Menus:
`Notice: Undefined index: id in ROOT/administrator/components/com_associations/models/fields/itemlanguage.php on line 93`
Categories:
`Warning`
`Error loading component: category, Component not found.`
`Warning: Invalid argument supplied for foreach() in ROOT/libraries/cms/helper/content.php on line 119`
Articles: we get the articles manager in the left part
- `You are not permitted to use that link to directly access that page (#1).`
  (article is checked out)

administrator/components/com_associations/views/association/tmpl/edit.php
andrepereiradasilva: you should not use the modal layout (that was just a PoC) you need to use the edit normal layout of the component item in the side by side.

administrator/components/com_associations/views/association/tmpl/edit.php
andrepereiradasilva: hum .. for me the question is more: why are you reloading the whole page when, IMHO you need to just reload the right iframe?
anyway, i can be missing something.

administrator/components/com_associations/views/association/view.html.php
```diff
@@ -139,9 +143,9 @@ protected function addToolbar()
 		JToolbarHelper::apply($associatedView . '.apply');
 		JToolbarHelper::save($associatedView . '.save');
 		JToolbarHelper::save2new($associatedView . '.save2new');
-		JToolbarHelper::cancel($associatedView . '.cancel', 'JTOOLBAR_CLOSE');
+		JToolbarHelper::cancel('association.cancel', 'JTOOLBAR_CLOSE');
```
andrepereiradasilva: nice. the Close button works fine now!
But attention, you need to checkout the items too on close.

administrator/components/com_associations/views/association/tmpl/edit.php
jreys: so, use the normal layout and hide the toolbar? because the normal layout brings the page's header

andrepereiradasilva: normal but with component template `&tpml=component` i think

administrator/components/com_associations/views/association/view.html.php
```diff
@@ -93,32 +93,36 @@ public function display($tpl = null)
...
+
+		$associatedComponent = $input->get('acomponent', '', 'string');
+		$associatedView      = $input->get('aview', '', 'string');
+
+		$options = array(
+			'option'    => $associatedComponent,
+			'view'      => $associatedView,
+			'task'      => $associatedView . '.edit',
+			'layout'    => 'edit',
+			'tmpl'      => 'component',
```
jreys, line 110: @andrepereiradasilva it already is tmpl = component, what I meant was, how do I save the item without loading the whole thing? i.e., save + load with tmpl=component. Should I save and reload the iframe? is this a good approach?

andrepereiradasilva, line 110: i don't see why not.
@infograf768

administrator/components/com_associations/views/association/tmpl/edit.php
alikon: can be done with some ajax call ?

alikon:
testing with articles category  KO
gsoccategory

alikon:
testing with menu that already have associotians KO
menugsoc

alikon:
testing contact category ko , category is buggy
contactcat

jreys:
yep, category is implemented the same way for every item, so no need to test more categories, it's broken :stuck_out_tongue:

alikon:
article and conatc seems to work so category is broken , yes...

administrator/components/com_associations/views/association/tmpl/edit.php
jreys: already fixed @alikon

alikon:
testing menu that already have associations ko
menui

alikon:
@jreys 
testing menu that already have associations is still  ko

menu2

alikon:
on contact categories what is the user 42 warning (on the left side) ??? i don't have such userid

warningcategory

infograf768:
Applied
https://github.com/joomla/joomla-cms/pull/11220
and then tested.

A lot of improvements!
We still have the check-in issue, i.e. when getting on the side by side and Closing (for example a categories side by side), both categories displayed are checked-out.

For Menus, I get an error with 11220 applied
Notice: Undefined variable: input in ROOT/administrator/components/com_menus/views/item/tmpl/edit.php on line 80
See my comment here:
https://github.com/joomla/joomla-cms/pull/11220#issuecomment-234163944
- the usual Undefined index:id

infograf768:
Also, when using the Close button I get:
[21-Jul-2016 06:04:41 UTC] PHP Strict standards:  Declaration of AssociationsControllerAssociation::cancel() should be compatible with JControllerForm::cancel($key = NULL) in ROOT/administrator/components/com_associations/controllers/association.php on line 18

infograf768:
concerning the undefined index:id, I guess this would solve it:

```
            if (isset($associations[$lang->value]))
            {
                if ($associatedComponent != 'com_menus')
                {
                    parse_str($associations[$lang->value], $contents);
                    $removeExtra  = explode(":", $contents['id']);
                    $lang->value  = $removeExtra[0];
                }
                else
                {
                    $lang->value = $associations[$lang->value];
                }
            }
            else
            {
                $lang->value = 0;
            }
```

andrepereiradasilva:
@alikon for the warning please check https://github.com/joomla/joomla-cms/issues/11228 and see if you can reproduce

alikon:
I'll check it

jreys:
@infograf768 , had to do it like this:

```
if ($lang->value == $forcedLanguage)
{
    unset($existingLanguages[$key]);
}
if (isset($associations[$lang->value]))
{
    if ($associatedComponent != 'com_menus')
    {
        parse_str($associations[$lang->value], $contents);
        $removeExtra  = explode(":", $contents['id']);
        $lang->value  = $removeExtra[0];
    }
    else
    {
        $lang->value = $associations[$lang->value];
    }
}
else
{
    $lang->value = 0;
}
```

Otherwise the reference language would be on the target and we don't want that.

administrator/components/com_associations/views/association/tmpl/edit.php
```diff
@@ -33,7 +26,35 @@ function triggerSave() {
...
 		});
+
+		$('.btn-success').attr('onclick','return triggerSave()');
+
+		$('#reference-association').load(function (){
+			$('#reference-association').each(function () {
+				$(this).contents().find('.chzn-single').css('background', 'transparent');
+				$(this).contents().find('.chzn-single').css('background-color', '#eee');
+				$(this).contents().find('.controls').css( 'pointer-events', 'none' );
+    			$(this).contents().find('input').attr('disabled', 'disabled');
```
andrepereiradasilva: in this case. i don't know if readonly or disabled
Anyway the tinymce has the readonly too. something like this
```editorobj.setMode('readonly');``` or ```readonly: true,``` on init
Also check the "Permissions" tab, those are not blocked.

But thinking out loud, maybe the user sees an error in the reference language and want to edit it too... maybe it should be editable too if needed. In that case another save button is needed. what do you think?

Comment (jreys) [inline administrator/components/com_associations/views/association/tmpl/edit.php]: I think you're right, maybe put the "save" triggering both iframes?

Comment (andrepereiradasilva) [inline administrator/components/com_associations/views/association/tmpl/edit.php]: i would recommend that as i certanly can see the use for that, but let's see @infograf768 opinion on this.

Comment (jreys) [inline administrator/components/com_associations/views/association/tmpl/edit.php]: I'm thinking that as an ACL option, imagine having an admin that wants his users to translate some items but wants to keep the default language intact.

Comment (andrepereiradasilva) [inline administrator/components/com_associations/views/association/tmpl/edit.php]: i see interesting idea.

so like a `core.edit.reference` ?

Comment (jreys) [inline administrator/components/com_associations/views/association/tmpl/edit.php]: either that or blocking the edition of a certain language like it was originally planned, both options sound good :smile:

File: administrator/components/com_associations/views/association/tmpl/edit.php [hunk 12]
```diff
@@ -14,15 +14,8 @@
...
-	function iframeRef( frameRef ) {
-	return frameRef.contentWindow
-		? frameRef.contentWindow.document
-		: frameRef.contentDocument
-	}
-
 	function triggerSave() {
-		var inside = iframeRef( document.getElementById('target-association') );
-		inside.getElementById('applyBtn').click();
+		window.frames[1].Joomla.submitbutton('" . $this->associatedView . ".apply');
```
andrepereiradasilva: we never now if custom 3º party admin modules add more frames to the window.
the more save option is to get the iframe object by unique id

infograf768:
Tested some:

I chose a menu item which already had associations.
I modified stuff in the Target and Saved.

Result:
The Target item IS saved and it is also checked-in. 😃 
(Check-in is not working for Articles or categories)
But any other Target I just displayed, as well as the Reference item, are checked-out.

infograf768:
Other aspects:
1. When using the save button (for target or Reference (although reference can't be edited for the moment). We get admin menus in the Iframe.
2. The Target language field should always be read-only and filled corresponding to the lang Selector
3. When saving a NEW target for which there was no association before, the item is not associated.

andrepereiradasilva:
@infograf768 

> 1. When using the save button (for target or Reference (although reference can't be edited for the moment). We get admin menus in the Iframe.

Use latest staging. this is already solved.

> 2. The Target language field should always be read-only and filled corresponding to the lang Selector

Agree! I also think the same should happen for the reference language iframe!

> 3. When saving a NEW target for which there was no association before, the item is not associated.
 Show all checks

Confirmed. This is one of the main goals of this component ....

Another thing. About the `- Select language -` box behaviour.

IMO the `- Select language -` select box should have the list of all langauges (except the reference language). ^This is how it works right now.
But, the behaviour should be different:
- If the user select `- Select language -` if should do no action and go back to the current iframe selected language.
- If the user select a language that  **doesn't exist a item with association** if should open a new item with the selected language preselected and the field disabled.
- If the user select a language that **already exist a item with association** if should open the edit item with the selected language preselected and the field disabled.

infograf768:
> Use latest staging. this is already solved.

Yep but check-in is not done yet

andrepereiradasilva:
still much to do ... i'm afraid

andrepereiradasilva:
BTW @jreys in case you are wondering how to create associations

See:
- Main Admin model https://github.com/joomla/joomla-cms/blob/staging/libraries/legacy/model/admin.php#L1211-L1262
- Menu item override https://github.com/joomla/joomla-cms/blob/staging/administrator/components/com_menus/models/item.php#L1378-L1451
- Category override https://github.com/joomla/joomla-cms/blob/staging/administrator/components/com_categories/models/category.php#L570-L642

infograf768:
in association/view.html.php
After 

```
    protected function addToolbar()
    {
        $jinput = JFactory::getApplication()->input;
```

Add
`$jinput->set('hidemainmenu', 1);`

in order to not display the admin menus as we are in an edit page.

[EDIT]
I gess you can use `$input` and not `$jinput` there

infograf768:
REFERENCE
Result of saving reference (here an fr-FR article) 
screen shot 2016-07-24 at 08 56 13
- the change in the text-area is not saved. (Nothing else is editable.)

TARGET
Saving Target when it does not exist:
Language is readonly OK
Item is not automatically associated
Associations tab displays=> it should not.
=>Note: once Target saved, it is still possible to edit it and associate OK. Saving again does not update the associations for the Reference in the side by side, but they show OK when displaying the Article Manager.

alikon:
article side by side js error

jserror

infograf768:
To avoid the js error, only display the Iframe when `selectedLang` has a value, it will also avoid saving anything when no language is selected (also: would be good to display `Save Target` only when `selectedLang` is not empty.)
after

```
            selectedLang = $('#jform_itemlanguage').val();
```

something like

```
            if (selectedLang == '')
            {
                $('#target-association').hide();
            }
            else
            {
                $('#target-association').show();
            }
```

result:

screen shot 2016-07-25 at 10 53 46

administrator/components/com_associations/controllers/association.php
```diff
@@ -27,6 +27,20 @@ class AssociationsControllerAssociation extends JControllerForm
...
+		$extension = $this->input->get('extension', '', 'string');
+		$refID     = $this->input->get('id', '', 'int');
+
+		$getCP = $extension != '' ? ('com_categories.category|' . $extension) : ($component . '.' . $view);
+
+		$checkOutComponent = AssociationsHelper::getComponentProperties($getCP);
+
+		if (!is_null($checkOutComponent->fields->checked_out))
+		{
+			$checkOutComponent->table->checkin(4);
```
andrepereiradasilva: what is this harcoded "4" ??

also you need to make two checkin, one for reference, one for target.
in other words you need to do a javascript like cancel(array(idreference, idtarget)); (js equivalent)
then process that `$key` in cancel method to checkin both.

alikon:
Test case: Articles Category,
select a Reference category (en-GB) an then select as Target the already associated category it-IT

modify same text on both side and save Reference and after save Target

errormc

infograf768:
@test
Great improvements! 👍 

Remains to do:
Saving a new target item for a language does not associate it automatically.

I know ACL are still to be implemented but at least the language in the reference item should never be modified.
i.e. after 

```
        $('#reference-association').load(function (){
            //Disable reference fields, commented out just to keep the code for ACL
            $('#reference-association').each(function () {
[...]
```

should be added in any case:

```
$(this).contents().find('#jform_language_chzn').css( 'pointer-events', 'none' );
                $(this).contents().find('#jform_language_chzn').find('.chzn-single').css('background', 'transparent');
                $(this).contents().find('#jform_language_chzn').find('.chzn-single').css('background-color', '#eee');
```

When the reference item is saved, the whole page has to be reloaded in order to update the possible targets (important when an association has been modified in the reference item).

Copy reference to Target for the moment (articles) just copies the title.

infograf768:
I tested this concerning the reference item minimum accepted modifications. It lets modify Associations and prevents changing anyhting else (left some stuff commented).

```
        $('#reference-association').load(function (){
            //Disable reference fields, commented out just to keep the code for ACL
            $('#reference-association').each(function () {
                //$(this).contents().find('.chzn-single').css('background', 'transparent');
                //$(this).contents().find('.chzn-single').css('background-color', '#eee');
                $(this).contents().find('.controls').css( 'pointer-events', 'none' );
                $(this).contents().find('.editor').css( 'pointer-events', 'none' );
                $(this).contents().find('select').css( 'pointer-events', 'none' );
                //$(this).contents().find('input').attr('disabled', 'disabled');
                $(this).contents().find('#jform_language_chzn').css( 'pointer-events', 'none' );
                $(this).contents().find('#jform_language_chzn').find('.chzn-single').css('background', 'transparent');
                $(this).contents().find('#jform_language_chzn').find('.chzn-single').css('background-color', '#eee');
                $(this).contents().find('#associations .controls').css( 'pointer-events', 'auto' );
            });
            $('#toolbar-copy').children().first().attr('onclick', 'return copyRefToTarget()');
        });
```

[EDIT]
`$(this).contents().find('.editor').css( 'pointer-events', 'none' );`
would prevent copy/paste from editor. Not so good.

infograf768:
As ACL is quite complex here, I wonder if we should not define minimum user group general permissions (not only the specific com_associations permissions) to be able to use the component. For example: the group should be able to checkin, edit and create in all components.

andrepereiradasilva:
> As ACL is quite complex here, I wonder if we should not define minimum user group general permissions (not only the specific com_associations permissions) to be able to use the component. For example: the group should be able to checkin, edit and create in all components.

For me ACL, as a start point, ACL would be basicly the components ACL.
already added an example on how to do that in https://github.com/joomla-projects/gsoc16_improved-multi-lingual/issues/7#issuecomment-234711420

infograf768:
I explain why we may have issues with this:
In present staging, one can have access to an item but not to a possible associated item and still be able to associate them through editing the item to which one has edit rights.

Wrong or not, If we implement full components ACL in com_associations, we change this present behaviour. 

If we change it in com_associations, then it should be changed all over the place, thus my simpler proposal above.

infograf768:
that's why I propose:
Associating items should be a specific permission, simple to implement:
One should have create/edit rights for the specific component (not all components indeed).

If not such permissions, then the list should not propose the items to associate in com_associations.

jreys:
@infograf768 , I tried saving a new item and it saves the association, can you tell me how you tested it? My guess is, you saved the target several times and then you saved the reference, I'll take a look at that.

administrator/components/com_associations/views/association/tmpl/edit.php
```diff
@@ -14,15 +14,14 @@
...
-	function triggerSave() {
-		var inside = iframeRef( document.getElementById('target-association') );
-		inside.getElementById('applyBtn').click();
+	function triggerSave(frame) {
+		if (frame == 'reference') {
+			window.frames[0].Joomla.submitbutton('" . $this->associatedView . ".apply');
+		}
+		if (frame == 'target') {
+			window.frames[1].Joomla.submitbutton('" . $this->associatedView . ".apply');
+			window.frames[0].location.reload();
```
andrepereiradasilva: ok... so what if i add changed any of the reference language fields? will i lost all my changes, right?
So you have to options. 
1. Intead of reloading, add the reference dinamycly to the reference iframe associations tab. 
2. Use reload, but use a confirm dialog or something here.

My vote clearly  goes to 1. because of UX.

administrator/components/com_associations/views/association/tmpl/edit.php
```diff
@@ -20,7 +20,16 @@ function triggerSave(frame) {
...
 			window.frames[1].Joomla.submitbutton('" . $this->associatedView . ".apply');
-			window.frames[0].location.reload();
+
+			jQuery('#target-association').load(function () {
+				target = jQuery('#reference-association').contents();
+				selectedLang = jQuery('#jform_itemlanguage').val();
+				split = selectedLang.split('|');
+				langAssociation = split[0].replace('-','_');
+				langID = jQuery(this).contents().find('#jform_id').val();
+				console.log(target.find('#jform_associations_' + langAssociation + '_id').html());
```
andrepereiradasilva: console.log ?

andrepereiradasilva:
i will remove all my already solved comment from this PR to make it more light

administrator/components/com_associations/views/association/view.html.php
```diff
@@ -93,32 +93,41 @@ public function display($tpl = null)
...
-		$referenceId = $input->get('id', '0');
-		$associatedComponent = $input->get('acomponent', '');
-		$associatedView = $input->get('aview', '');
-
-		$this->link = "";
-
-		if ($associatedComponent == 'com_categories')
+		$this->form  = $this->get('Form');
+		$input     = $this->app->input;
+		$assoc     = JLanguageAssociations::isEnabled();
```
andrepereiradasilva: this can be removed. not used

administrator/components/com_associations/views/association/tmpl/edit.php
```diff
@@ -63,35 +180,39 @@ function triggerSave() {
...
+		width: 50%;
+	}
 ');
 ?>
-<button id="toogle-left-panel">Show/Hide Reference (PoC)</button>
+<button id="toogle-left-panel" class="btn btn-small">Show/Hide Reference (PoC)</button>
 
-<form action="<?php echo JRoute::_('index.php?option=com_associations&view=association'); ?>"
- method="post" name="adminForm" id="item-form" class="form-validate">
+<form action="<?php echo JRoute::_(JFactory::getURI()->toString()); ?>"
```
andrepereiradasilva: it's better to add the all URL parameters inside JRoute, like all other views.
We need always to be consistent with the rest of the code in the core.

administrator/components/com_associations/views/association/tmpl/edit.php
```diff
@@ -14,15 +14,23 @@
...
-	return frameRef.contentWindow
-		? frameRef.contentWindow.document
-		: frameRef.contentDocument
-	}
+	function triggerSave(frame) {
+		if (frame == 'reference') {
+			window.frames[0].Joomla.submitbutton('" . $this->associatedView . ".apply');
+		}
+		if (frame == 'target') {
+			window.frames[1].Joomla.submitbutton('" . $this->associatedView . ".apply');
```
andrepereiradasilva: we never know if custom 3º party admin modules add more frames to the window.
the more save option is to get the iframe object by unique id, not by frames array fixed key

administrator/components/com_associations/views/association/tmpl/edit.php
andrepereiradasilva: BTW save still doe not work

administrator/components/com_associations/models/association.php
```diff
@@ -16,5 +16,27 @@
...
+	 * Method to get the record form.
+	 *
+	 * @param   array    $data      Data for the form.
+	 * @param   boolean  $loadData  True if the form is to load its own data (default case), false if not.
+	 *
+	 * @return  mixed  A JForm object on success, false on failure
+	 *
+	 * @since   __DEPLOY_VERSION__
+	 */
+	public function getForm($data = array(), $loadData = true)
```
andrepereiradasilva: for what is this used?

jreys: $this->form->getInput('itemlanguage');
here: https://github.com/joomla-projects/gsoc16_improved-multi-lingual/pull/17/files#diff-d7a25844a430646eaa1f3071a83a58e0R209

if there's an easier way please do tell :smile:

jreys:
@alikon , can you explain better your test please? what text fields did you change and to which values? if I can do a similar test, I can find that error, currently as you said it I had no problems :worried:

administrator/components/com_associations/helpers/associations.php
```diff
@@ -83,6 +83,16 @@ public static function getComponentProperties($component = '')
...
+			
+			if (isset($properties->tableFields['created_by']))
+			{
+				$properties->fields->created_by  = 'created_by';
+			}
+			else
+			{
+				$properties->fields->created_by = isset($properties->tableFields['created_user_id']) ? 'created_user_id' : null;
+			}
+
```
andrepereiradasilva: please follow the same logic as the others
```php
$properties->fields->created_by   = isset($properties->tableFields['created_user_id']) ? 'created_user_id' : null;
$properties->fields->created_by   = isset($properties->tableFields['created_by']) ? 'created_by' : $properties->fields->created_by;
```

administrator/components/com_associations/models/association.php
andrepereiradasilva: yes ... add association.xml to /model/forms/association.xml and it will auto load the form
see how other components do it.

administrator/components/com_associations/models/associations.php
```diff
@@ -121,8 +121,14 @@ protected function getListQuery()
 		$query->select($db->quoteName('a.id'))
 			->select($db->quoteName('a.' . $component->fields->title, 'title'))
 			->select($db->quoteName('a.' . $component->fields->alias, 'alias'))
-			->select($db->quoteName('a.' . $component->fields->ordering, 'ordering'))
-			->from($db->quoteName($component->dbtable, 'a'));
+			->select($db->quoteName('a.' . $component->fields->ordering, 'ordering'));
+
+		if ($component->dbtable != '#__menu')
```
andrepereiradasilva: put a comment above explaining please. see the other ifs

administrator/components/com_associations/models/associations.php
```diff
@@ -121,8 +121,14 @@ protected function getListQuery()
...
+		{
+			$query->select($db->quoteName('a.' . $component->fields->created_by, 'created_by'));
+		}
+		
+		$query->from($db->quoteName($component->dbtable, 'a'));
```
andrepereiradasilva: please leave the for where it was before.

infograf768:
After correcting some of the stuff noted above by Andre and applying PR https://github.com/joomla-projects/gsoc16_improved-multi-lingual/pull/26
I can save a new item as associated. But it still remains checked out!
Also, it does not update the associations tab in the reference item in the side by side view.

infograf768:
New item now checked-in OK.
New issue:
After saving this new item, then selecting another language and then selecting again the language of the new item, the form is empty, as if there was no associations defined for that language. One has to Close, then click again on the reference item to get the correct display for the new associated item.

infograf768:
https://github.com/joomla-projects/gsoc16_improved-multi-lingual/pull/17/commits/0da868adb32e83c7eac3c735e9b74192fca7a38c solves that issue :)

administrator/components/com_associations/helpers/associations.php
andrepereiradasilva: forget it. i already added this in my merged PR, please remove this changres so you don't get conflicts

administrator/components/com_associations/views/associations/tmpl/default.php
```diff
@@ -83,8 +85,10 @@
 				</tr>
 			</tfoot>
 			<tbody>
-			<?php foreach ($this->items as $i => $item) : ?>
-				<?php // @todo ACL check ?>
+			<?php foreach ($this->items as $i => $item) :
+				$canEdit    = $user->authorise('core.edit', $this->state->get('component') . $item->id);
+				$canEditOwn = $user->authorise('core.edit.own', $this->state->get('component') . $item->id) && $item->created_by == $userId;
```
andrepereiradasilva: - Does this work for a component that does not have this ACL (`core.edit`, 'core.edit.own') permissions?
- Does this work for a component that does not have item level permissions (ex: com_menus), i.e., just permissions?
- Isn't needed to have more permissions checked? Other components check more things.

administrator/components/com_associations/views/association/view.html.php
```diff
@@ -132,14 +141,15 @@ public function display($tpl = null)
...
+		$input->set('hidemainmenu', 1);
+
 		JToolbarHelper::title(JText::_('COM_ASSOCIATIONS_HEADER_EDIT'), 'contract');
 
-		JToolbarHelper::apply($associatedView . '.apply');
-		JToolbarHelper::save($associatedView . '.save');
-		JToolbarHelper::save2new($associatedView . '.save2new');
-		JToolbarHelper::cancel($associatedView . '.cancel', 'JTOOLBAR_CLOSE');
+		JToolbarHelper::apply('reference', 'COM_ASSOCIATIONS_SAVE_REFERENCE');
+		JToolbarHelper::apply('target', 'COM_ASSOCIATIONS_SAVE_TARGET');
```
andrepereiradasilva, line 146: compoent ACL check is needed here
see how other components does it

administrator/components/com_associations/views/association/tmpl/edit.php
```diff
@@ -63,35 +198,39 @@ function triggerSave() {
...
 	<div class="outer-panel" id="right-panel">
 		<div class="inner-panel">
-			<h3><?php echo JText::_('COM_ASSOCIATIONS_ASSOCIATED_ITEM'); ?></h3>
-			<select id="ref-language" name="ref-language">
-				<?php echo JHtml::_('select.options', JHtml::_('contentlanguage.existing', false, true), 'value', 'text'); ?>
-			</select>
-			<button onclick="return triggerSave();" class="btn btn-small btn-success">
-				<span class="icon-apply icon-white"></span>Save
-			</button>
+			<div class="reference-side">
```
andrepereiradasilva: why add a new div and css class here? Do we really need to?

administrator/components/com_associations/views/association/tmpl/edit.php
```diff
@@ -63,35 +198,39 @@ function triggerSave() {
...
- method="post" name="adminForm" id="item-form" class="form-validate">
+<form action="<?php echo JRoute::_(JFactory::getURI()->toString()); ?>"
+ method="post" name="adminForm" id="adminForm" class="form-validate">
 
 <div class="sidebyside">
 	<div class="outer-panel" id="left-panel">
 		<div class="inner-panel">
 			<h3><?php echo JText::_('COM_ASSOCIATIONS_REFERENCE_ITEM'); ?></h3>
-			<iframe src="<?php echo JRoute::_($this->link); ?>" 
+			<iframe id="reference-association" src="<?php echo JRoute::_($this->link); ?>" class="reference-association"
```
andrepereiradasilva: where is this  `class="reference-association"` ever used?

administrator/components/com_associations/views/association/tmpl/edit.php
```diff
@@ -63,35 +198,39 @@ function triggerSave() {
...
 				name="<?php echo JText::_('COM_ASSOCIATIONS_TITLE_MODAL'); ?>" height="100%" width="400px" scrolling="no">
```
andrepereiradasilva: COM_ASSOCIATIONS_TITLE_MODAL ??

administrator/components/com_associations/views/association/tmpl/edit.php
```diff
@@ -63,35 +198,39 @@ function triggerSave() {
...
-				<span class="icon-apply icon-white"></span>Save
-			</button>
+			<div class="reference-side">
+				<h3><?php echo JText::_('COM_ASSOCIATIONS_ASSOCIATED_ITEM'); ?></h3>
+				<?php echo $this->form->getInput('itemlanguage'); ?>
+			</div>
 			<iframe id="target-association" name="target-association" 
-				src="<?php echo JRoute::_($this->link); ?>" 
+				src="<?php echo JRoute::_($this->targetLink); ?>" 
 				name="<?php echo JText::_('COM_ASSOCIATIONS_TITLE_MODAL'); ?>" height="100%" width="400px" scrolling="no"></iframe>
```
andrepereiradasilva: COM_ASSOCIATIONS_TITLE_MODAL ??

administrator/components/com_associations/models/fields/itemlanguage.php
```diff
@@ -0,0 +1,113 @@
...
+			JLoader::register($helpAssoc, JPATH_SITE . '/components/' . $associatedComponent . '/helpers/association.php');
+			$helpRoute = str_replace('com_', '', $associatedComponent . 'HelperRoute');
+			JLoader::register($helpRoute, JPATH_SITE . '/components/' . $associatedComponent . '/helpers/route.php');
+			if (class_exists($helpAssoc) && is_callable(array($helpAssoc, 'getAssociations')))
+			{
+				$associations = call_user_func(array($helpAssoc, 'getAssociations'), $referenceId, $associatedView);
+				$existingLanguages = JHtml::_('contentlanguage.existing', false, true);
+			}
+			
+		}
```
andrepereiradasilva: I think you can now replace now all this if/elseif/else by:
```php
$key = $extension !== '' ? 'com_categories.category|' . $extension : $component . '.' . $view;
$cp  = AssociationsHelper::getComponentProperties($key);

JLoader::register($cp->associations->gethelper->class, $cp->associations->gethelper->file);
$associations      = call_user_func(array($cp->associations->gethelper->class, $cp->associations->gethelper->method), $referenceId, $associatedView);
$existingLanguages = JHtml::_('contentlanguage.existing', false, true);
```

But please test.

jreys: As it is, it won't work, the layout is fine but the associated items are not loaded, small test:

```
<select id="jform_itemlanguage" name="jform[itemlanguage]" onchange="loadFrame(jQuery('#jform_itemlanguage').val());" class="chzn-done" style="display: none;">
	<option value="" selected="selected">- Select Target Language -</option>
	<option value="fr-FR|0">French (FR)</option>
	<option value="it-IT|0">Italian (IT)</option>
	<option value="pt-PT|0">Português (pt-PT)</option>
</select>
```

administrator/components/com_associations/views/association/tmpl/edit.php
```diff
@@ -33,7 +59,116 @@ function triggerSave() {
...
+		//Save button action, replacing the fault Joomla.submitbutton() with custom function
+		$('.btn-success').attr('onclick', function(){
+			var frame = $(this).attr('onclick');
+
+			frame = frame.substring(frame.lastIndexOf('(')+1,frame.lastIndexOf(')'));
+			return 'return triggerSave(' + frame + ')';
+		});
+
+		$('#reference-association').load(function (){
+			//Disable reference fields, commented out just to keep the code for ACL
```
andrepereiradasilva: to disable a chosen field you should use chosen api
exmaple:
```js
$('#foobar').prop('disabled', true).trigger("liszt:updated");
```

administrator/components/com_associations/models/fields/itemlanguage.php
andrepereiradasilva: ok so works for menus
on categories we get an empty list
on others we get and helper route error???

jreys: I'm on it, the helper route error is normal since this code is missing:
```
$helpRoute = str_replace('com_', '', $associatedComponent . 'HelperRoute');
JLoader::register($helpRoute, JPATH_SITE . '/components/' . $associatedComponent . '/helpers/route.php');
```

andrepereiradasilva: works for others now
we need to change this is the core for work. https://github.com/joomla-projects/gsoc16_improved-multi-lingual/pull/28

only left fixing the categories

administrator/components/com_associations/models/associations.php
```diff
@@ -124,6 +124,12 @@ protected function getListQuery()
 			->select($db->quoteName('a.' . $component->fields->ordering, 'ordering'))
 			->from($db->quoteName($component->dbtable, 'a'));
 
+		// Select author for ACL checks
+		if ($component->dbtable != '#__menu')
```
andrepereiradasilva: why menu? what if other third party components doesn't sipport checkout?
Isn't it better ...
`!is_null($component->fields->created_by)`

administrator/components/com_associations/models/fields/itemlanguage.php
```diff
@@ -0,0 +1,82 @@
...
+		$referenceId         = $input->get('id', '');
+		$associatedComponent = $input->get('acomponent', '');
+		$associatedView      = $input->get('aview', '');
+		$extension           = $input->get('extension', '');
+		$forcedLanguage      = $input->get('forcedlanguage', '');
+
+		$key = $extension !== '' ? 'com_categories.category|' . $extension : $associatedComponent . '.' . $associatedView;
+		$cp  = AssociationsHelper::getComponentProperties($key);
+		$associations      = call_user_func(array($cp->associations->gethelper->class, $cp->associations->gethelper->method), $referenceId, $associatedView);
+		$existingLanguages = JHtml::_('contentlanguage.existing', false, true);
```
andrepereiradasilva: Atention. this still doesn't work for categories.
Have to check why.

administrator/components/com_associations/views/association/tmpl/edit.php
jreys: If you can tell me a way to have the filter and the label "Target" in the same line without doing this please tell me :smiley:

administrator/components/com_associations/views/association/tmpl/edit.php
jreys: I've had problems before with chosen fields:

```js
$(this).contents().find('#jform_language').prop('disabled', true).trigger('liszt:updated');
```

Result:
```html
<select id="jform_language" name="jform[language]" class="chzn-done" disabled="" style="display: none;">
[...]
</select>
```

But the update is not being made, don't know why

administrator/components/com_associations/views/association/tmpl/edit.php
```diff
@@ -33,7 +59,116 @@ function triggerSave() {
...
+					else {
+						$('#target-id').val($('#target-id').val()+','+split[1]);
+					}
+				}
+
+				//Auto-picking the selected language on the switcher
+				target.contents().find('#jform_language option[value='+selectedLang+']').attr('selected','selected');
+				target.contents().find('#jform_language').chosen();
+				target.contents().find('#jform_language').trigger('liszt:updated');
+				target.contents().find('#jform_language').parent().find('div:gt(2)').remove();
```
jreys: @andrepereiradasilva look at these lines, this must be wrong because I had to hardcode it to perform the behaviour I needed:
Line 120 - I find the '#jform_language' value
Line 121 - Don't know why, it works
Line 122 - Trigger update
Line 123 - I remove an extra language field that was added with .chosen(), if I remove line 121 all these lines are useless

administrator/components/com_associations/views/associations/tmpl/default.php
andrepereiradasilva:
```
Notice: Undefined property: stdClass::$created_by in /path/to/joomla-gsoc/administrator/components/com_associations/views/associations/tmpl/default.php on line 88
Notice: Undefined property: stdClass::$created_by in /path/to/joomla-gsoc/administrator/components/com_associations/views/associations/tmpl/default.php on line 88
```

administrator/components/com_associations/views/association/tmpl/edit.php
andrepereiradasilva: i'm no js expert, but something doesn't seem quite right here

andrepereiradasilva: when i hear `Line 121 - Don't know why, it works` it sign somethng is not good.

administrator/components/com_associations/models/association.php
```diff
@@ -16,5 +16,27 @@
...
+		// Get the form.
+		$form = $this->loadForm('com_associations.association', 'association', array('control' => 'jform', 'load_data' => $loadData));
+		if (empty($form))
+		{
+			return false;
+		}
+		else
+		{
+			return $form;
+		}
```
andrepereiradasilva: suggestion simplify
````php

return !empty($form) ? $form : false;
```

administrator/components/com_associations/models/fields/itemlanguage.php
```diff
@@ -0,0 +1,84 @@
...
+					$lang->value  = $lang->value . "|" . $removeExtra[0];
+				}
+				else
+				{
+					$lang->value = $lang->value . "|" . $associations[$lang->value];
+				}
+			}
+			else
+			{
+				$lang->value = $lang->value . "|" . 0;
```
andrepereiradasilva: Simplify. Do you mean?
```php
$lang->value .= '|0';
```

administrator/components/com_associations/views/association/tmpl/edit.php
jreys: why would a 3rd-party extension want that? and even if they do, all their frames would be inside the main iframes and this code won't reach them

administrator/components/com_associations/views/associations/tmpl/default.php
```diff
@@ -85,7 +85,10 @@
 			<tbody>
 			<?php foreach ($this->items as $i => $item) :
 				$canEdit    = $user->authorise('core.edit', $this->state->get('component') . $item->id);
-				$canEditOwn = $user->authorise('core.edit.own', $this->state->get('component') . $item->id) && $item->created_by == $userId;
+				if (isset($item->created_by))
```
andrepereiradasilva: `!is_null` not `isset`

andrepereiradasilva: sorry

administrator/components/com_associations/views/association/tmpl/edit.php
andrepereiradasilva: no, why it would? i don't know. but it's possible to create an admin component and inject some kind of iframe in all admin pages,

i'm not saying that this a huge problem.
it's a good practice to not use array indexes in a flexible app like a cms is, when you can use direct id/name references.

administrator/components/com_associations/views/association/tmpl/edit.php
andrepereiradasilva: i will check that

administrator/components/com_associations/views/association/tmpl/edit.php
andrepereiradasilva: follow the same logic as  https://github.com/jreys/gsoc16_improved-multi-lingual/pull/6

administrator/components/com_associations/views/associations/tmpl/default.php
```diff
@@ -81,8 +83,13 @@
...
 			<tbody>
-			<?php foreach ($this->items as $i => $item) : ?>
-				<?php // @todo ACL check ?>
+			<?php foreach ($this->items as $i => $item) :
+				$canEdit    = $user->authorise('core.edit', $this->state->get('component') . $item->id);
+				if (isset($item->created_by))
+				{
+					$canEditOwn = $user->authorise('core.edit.own', $this->state->get('component') . $item->id) && $item->created_by == $userId;
+				}
+				?>
```
andrepereiradasilva, line 92: - Does this work for a component that does not have this ACL (`core.edit`, `core.edit.own`) permissions?
- Does this work for a component that does not have item level permissions (ex: com_menus), i.e., just component level ACL permissions?
- Isn't needed to have more permissions checked? Other components check more things.

administrator/components/com_associations/views/association/tmpl/edit.php
andrepereiradasilva: ok see https://github.com/jreys/gsoc16_improved-multi-lingual/pull/6

administrator/components/com_associations/views/association/tmpl/edit.php
andrepereiradasilva: let's leave that for the graphical review process then

alikon:
test scenario

i have 4 contact items one for each language (it,fr,gb,pt)  unassociated
so  i select the reference item in this example the GB one
then on reference go in tab associations and select an associations for PT for example 
and save reference

### **(is this the way  we can associate existing unassociated item?)**

but on the target side if i select the pt language it opens a new contatct item 
should be load the newly saved target pt item i think.....

but let's go on testing  and close, and then repoen the same GB contact item  and click on the associations tab and then on the target choose the PT lang

test_scene0

should not be visible the association tab on target....

jreys:
@alikon you're absolutely right, my solution for that you've already posted and @andrepereiradasilva also agrees it's the best way:

image

ideally when you click that button, it would load a modal with the language selected on the right, for example, english chosen on the right:

image

we would get this by pressing the button:

image

and after choosing, it would be loaded on the right with the association, pre-populated on a hidden field on the left.

also, @infograf768 , I would also like your opinion on this

infograf768:
If the button to select new association for a said language in Target is placed near to the Target Language Selector, I am fine with this solution.
Adding it to the toolbar makes the UX even more confusing imho

infograf768:
Changes for the helpers loader should NOT be included in this patch. They are already pesent in GSOC staging and Joomla staging

infograf768:
ALSO:
changing
`window.frames[0]` to `window.frames['reference-association']`
`window.frames[1]` to `window.frames['target-association']`

Breaks the Save functions in Firefox

administrator/components/com_associations/views/association/tmpl/edit.php
```diff
@@ -63,35 +203,48 @@ function triggerSave() {
...
 ?>
-<button id="toogle-left-panel">Show/Hide Reference (PoC)</button>
+<button id="toogle-left-panel" class="btn btn-small">Show/Hide Reference (PoC)</button>
 
-<form action="<?php echo JRoute::_('index.php?option=com_associations&view=association'); ?>"
- method="post" name="adminForm" id="item-form" class="form-validate">
+<form action="<?php echo JRoute::_('index.php?option=com_associations&view=association&layout=' . $layout . '&acomponent=' 
+		. $aComponent . '&aview=' . $aView . '&forcedlanguage=' . $fLanguage . '&id='
+		. $this->referenceID); ?>""
+ method="post" name="adminForm" id="adminForm" class="form-validate">
```
jreys: I've added the URL parameters as @andrepereiradasilva suggested, one thing here, the &forcedlanguage parameter is not being used anywhere, @infograf768 I think it was you who suggested this parameter, is it really needed?

administrator/components/com_associations/helpers/associations.php
```diff
@@ -85,7 +85,8 @@ public static function getComponentProperties($key = '')
 				// For categories.
 				else
 				{
-					$cp[$key]->associations->gethelper->class  = ucfirst(substr($cp[$key]->realcomponent, 4)) . 'HelperAssociation';
+					$catExtension = ucfirst(substr($cp[$key]->extension, 4));
+					$cp[$key]->associations->gethelper->class  = $catExtension . 'HelperAssociation';
 					$cp[$key]->associations->gethelper->method = 'getCategoryAssociations';
```
andrepereiradasilva: I tested witout this change and works fine in my test.
so please remove this change.

administrator/components/com_associations/views/association/tmpl/edit.php
```diff
@@ -33,7 +62,118 @@ function triggerSave() {
 			$('#left-panel').toggle();
 			$('#right-panel').toggleClass('full-width');
 		});
+
```
andrepereiradasilva: To preload the image in the loading layer, add, at the beggining of the document ready event

```js
// Create and append the loading layer.
Joomla.loadingLayer("load");
```

administrator/components/com_associations/views/association/tmpl/edit.php
```diff
@@ -33,7 +62,118 @@ function triggerSave() {
...
+		selectedLang = jQuery('#jform_itemlanguage').val();
+
+		if (selectedLang == '')
+		{
+			document.getElementById('target-association').src = '';
+		}
+		else
+		{
+			// Iframe load finished, hide Joomla loading layer.
+			Joomla.loadingLayer('show');
```
andrepereiradasilva: wrong comment, my fault, replace with
```js
// Iframe load start, show Joomla loading layer.
```

administrator/components/com_associations/views/association/tmpl/edit.php
andrepereiradasilva: for the direct links from the list view you will need a `targetLanguage` parameter.

example: direct link to edit article 13 and auto start with pt-PT language selected and iframe loaded.
`/administrator/index.php?option=com_associations&view=association&layout=edit&acomponent=com_ccontent&aview=article&extension=&id=13&targetLanguage=pt-PT`

this `targetLanguage` will them need to be processed in this side by side view.

So change the parameter to `targetLanguage` and leave that for later.

administrator/components/com_associations/views/association/view.html.php
```diff
@@ -93,32 +93,40 @@ public function display($tpl = null)
...
-
-		if ($associatedComponent == 'com_categories')
+		$this->form  = $this->get('Form');
+		$input     = $this->app->input;
+		
+		$formData = new JInput($input->get('jform', '', 'array'));
+
+		$associatedComponent = $input->get('acomponent', '', 'string');
+		$associatedView      = $input->get('aview', '', 'string');
+		$this->referenceID   = $input->get('id', 0, 'int');
```
andrepereiradasilva: code normalization: replace `referenceID` for `referenceId` where it is used.

administrator/components/com_associations/views/association/view.html.php
```diff
@@ -93,32 +93,40 @@ public function display($tpl = null)
...
-		$associatedComponent = $input->get('acomponent', '');
-		$associatedView = $input->get('aview', '');
-
-		$this->link = "";
-
-		if ($associatedComponent == 'com_categories')
+		$this->form  = $this->get('Form');
+		$input     = $this->app->input;
+		
+		$formData = new JInput($input->get('jform', '', 'array'));
```
andrepereiradasilva: where is this formData used?

administrator/components/com_associations/views/association/view.html.php
andrepereiradasilva, line 146: ok, this goes to ACL PR

administrator/components/com_associations/views/associations/tmpl/default.php
andrepereiradasilva, line 92: ok, this goes to ACL PR

components/com_contact/helpers/association.php
```diff
@@ -10,6 +10,7 @@
 defined('_JEXEC') or die;
```
andrepereiradasilva: this PR must not have changes to thios helpers.
They are already on gsoc staging. UPDATE your branches and this will disappear.

administrator/components/com_associations/views/association/tmpl/edit.php
```diff
@@ -63,35 +203,48 @@ function triggerSave() {
...
+
+$input      = $this->app->input;
+$layout     = $input->get('layout', '', 'string');
+$aComponent = $input->get('acomponent', '', 'string');
+$aView      = $input->get('aview', '', 'string');
+$fLanguage  = $input->get('forcedlanguage', '', 'string');
+
 ?>
-<button id="toogle-left-panel">Show/Hide Reference (PoC)</button>
+<button id="toogle-left-panel" class="btn btn-small">Show/Hide Reference (PoC)</button>
```
andrepereiradasilva: this needs to be in a language var.

jreys: I'm thinking of treating this in another PR (JS code revision) where I optimize my JS code and that Hide Show will be dynamic:
When reference is shown : Hide Reference
When referece is hidden : Show Reference

andrepereiradasilva: If so, add to the tasks list issue

administrator/components/com_associations/views/association/tmpl/edit.php
```diff
@@ -14,26 +14,167 @@
...
+		});
+
+		$('#target-association').load(function () {
+			target = $(this).contents();
+
+			//Hide associations tab	
+			//target.find('a[href=#associations]').parent().hide();
+
+			langAssociation = '" . str_replace('-', '_', $this->referenceLanguage) . "';
+			langID = " . $this->referenceID . ";
```
andrepereiradasilva: referenceId

administrator/components/com_associations/views/association/tmpl/edit.php
```diff
@@ -63,35 +204,48 @@ function triggerSave() {
...
+
 ?>
-<button id="toogle-left-panel">Show/Hide Reference (PoC)</button>
+<button id="toogle-left-panel" class="btn btn-small">Show/Hide Reference (PoC)</button>
 
-<form action="<?php echo JRoute::_('index.php?option=com_associations&view=association'); ?>"
- method="post" name="adminForm" id="item-form" class="form-validate">
+<form action="<?php echo JRoute::_('index.php?option=com_associations&view=association&layout=' . $layout . '&acomponent=' 
+		. $aComponent . '&aview=' . $aView . '&referencelanguage=' . $rLanguage . '&id='
+		. $this->referenceID); ?>""
```
andrepereiradasilva: referenceId

administrator/components/com_associations/views/association/view.html.php
```diff
@@ -93,32 +93,37 @@ public function display($tpl = null)
...
+		$this->referenceId   = $input->get('id', 0, 'int');
+
+		$options = array(
+			'option'    => $associatedComponent,
+			'view'      => $associatedView,
+			'extension' => '',
+			'task'      => $associatedView . '.edit',
+			'layout'    => 'edit',
+			'tmpl'      => 'component',
+			'id'        => $this->referenceID
```
andrepereiradasilva: referenceId

andrepereiradasilva:
let's wait for travis checks and them merge

PR #20, opened 2016-07-16
Improve list view

andrepereiradasilva:
A lot of improvements in the list view.

Done:
- do not hardcode component item db table anymore
- 5 new filters that display according to the componet selected
- Two new columns (state and acess level) that appears according to the componet selected
- only show items according to the view levels
- now besides the title, there is the alias and category (if component supports it)
- code review
- optimized code
- several minor changes
- minor bugs solved
- more extensible


infograf768:
Nice, except that when filtering by menu type (btw lang string missing), we get the mysql error when changing item to another item than menu.
Also, there are modifications to menus which should not be done here I guess.
I suggest you propose them to the CMS or find another way in com_associations.

andrepereiradasilva:
> Nice, except that when filtering by menu type (btw lang string missing), we get the mysql error when changing item to another item than menu.

will check

> Also, there are modifications to menus which should not be done here I guess.
> I suggest you propose them to the CMS or find another way in com_associations.

yeah, i have done those changes here or else it would produce a fatal errors. which would made impossible for testing.
but this is a cms bug on menus and user/group debug models.

andrepereiradasilva:
> (btw lang string missing)

Solved.

> Nice, except that when filtering by menu type (...), we get the mysql error when changing item to another item than menu.

i get no SQL error. The only SQL error i have is when messing with the ordering. As told in the PR description this is the remaining TO DO.
Try again but always using the default order: "Ordering ASC", don't chnage it for now.

andrepereiradasilva:
> Also, there are modifications to menus which should not be done here I guess.
> I suggest you propose them to the CMS or find another way in com_associations.

ok so i excluded com_users component from loading the model and just correct it here (temporary) for com_menus item model.

I proposed a PR to fix this in main joomla CMS.
See https://github.com/joomla/joomla-cms/pull/11164 and please test so it gets merged if ok.

andrepereiradasilva:
so if all ok with this i would get this merged and solve the ordering issue later.

infograf768:
The ordering issue is the same we discussed. Maybe I was unclear.
One gets a mysql error when displaying menu items, order set to Menus ordering and switching to another kind of items.
This was what I solved temporarily before with a setState.
Will test tomorrow what we have now.

andrepereiradasilva:
i know there is a SQL error with the ordering. that's why i told to test without changing the ordering.

> Try again but always using the default order: "Ordering ASC", don't chnage it for now.

I can fix the order easily to not produce SQL errors.

The thing is, i'm trying to think of a way to display the order options dinamycly according to the component selected. But haven't found one yet.

administrator/components/com_associations/models/fields/associatedcomponent.php
```diff
@@ -48,29 +73,30 @@ protected function getGroups()
...
+				if ($model && $model->get('associationsContext') && $model->get('typeAlias') && !in_array($model->get('typeAlias'), $typeAliasList))
+				{
 					// Load component language file.
+					$lang->load($component . '.sys', JPATH_ADMINISTRATOR) || $lang->load($component . '.sys', $componentAdminPath);
 					$lang->load($component, JPATH_ADMINISTRATOR) || $lang->load($component, $componentAdminPath);
 
-					// Add componet option select box.
-					$options[JText::_($component)][] = JHtml::_('select.option', $model->typeAlias, JText::_($component));
+					// Add component option select box.
+					$options[JText::_($component)][] = JHtml::_('select.option', $model->get('typeAlias'), JText::_($component));
```
andrepereiradasilva, line 97: for flexility with components with several item models we need to have a way to use a correct text variable that itentifies the item inside the component.

Proposal:

``` php
$options[JText::_($component)][] = JHtml::_('select.option', $model->get('typeAlias'), JText::_($component));
```

Should be

``` php
$options[JText::_($component)][] = JHtml::_('select.option', $model->get('typeAlias'), JText::_($component . '_ITEM_' . $itemName . 'S'));
```

For instance:
- COM_MENUS_ITEM_ITEMS="Menu items"
- COM_CONTENT_ITEM_ARTICLES="Articles"
- COM_NEWSFEEDS_ITEM_NEWSFEEDS="News Feeds"
- COM_CONTACT_ITEM_CONTACTS="Contacts"

@jreys @infograf768

andrepereiradasilva:
ok to not get more time on this, i solved the order SQL errors.

andrepereiradasilva:
@jreys @infograf768 can we merge this now?

infograf768:
This works fine. Ordering is working OK now. :clap:

I have another solution for the item strings (it would anyway be a requirement for 3rd party components. Notes should be added concerning all these requirements.)

What if we used for items:

`// Add component option select box.
                    $options[JText::_($component)][] = JHtml::_('select.option', $model->get('typeAlias'), JText::_(strtoupper($model->get('typeAlias'))));`

and for categories (**This would also solve the fact that we do not know after choosing a component categories in which component we are selecting**)

`$options[JText::_($component)][] = JHtml::_('select.option', 'com_categories.category|' . $component, JText::_(strtoupper($component . '.CATEGORIES')));`

(I know it is not really necessary to add `strtoupper`...)

Therefore adding these lang strings in core (a period in the constant is OK and these strings would stand up) in the component respective sys.ini:

```
COM_CONTENT.ARTICLE="Articles"
COM_CONTENT.CATEGORIES="Articles Categories"
COM_CONTACT.CONTACT="Contacts"
COM_CONTACT.CATEGORIES="Contacts Categories"
COM_MENUS.ITEM="Menu Items"
COM_NEWSFEEDS.NEWSFEED="Newsfeeds"
COM_NEWSFEEDS.CATEGORIES="Newsfeeds Categories"
COM_WEBLINKS.CATEGORIES="Weblinks Categories"
```

We would get:
screen shot 2016-07-18 at 08 33 03

For the moment I would add these strings in com_associations.sys.ini until com_associations is merged in core.

andrepereiradasilva:
ok so merge this as it it and made the changes later?

infograf768:
@jreys @andrepereiradasilva 
If you both agree, let's change the few things in my proposal and merge.

PR #24, opened 2016-07-25
Improve component load helper

andrepereiradasilva:
Load the model and the JTable in the helper.
Also checks for existence of checked_out


andrepereiradasilva:
after this PR to checkout you just need to do something like this. example:

``` php
$component = AssociationsHelper::getComponentProperties('com_categories.category|com_content');
if (!is_null($component->fields->checked_out))
{
    $component->table->checkOut(...);
}
```

PR #26, opened 2016-07-26
Generalizing hide associations

jreys:
#### Summary of Changes

@infograf768 , this is an improvement to your patch here:
https://github.com/joomla/joomla-cms/commit/e30ee0d14f8f4230ef1e6ff6a191f8fdedb5cf47#diff-57b1cb851829b953306d161dc65f7aa8R12, it's beeing done in one place rather than in several :smiley: 
#### Testing Instructions

Try editing a new item or an item that has it's language set to "*" -> All

### DIFF
--- a/layouts/joomla/edit/associations.php
+++ b/layouts/joomla/edit/associations.php
@@ -10,7 +10,7 @@
 defined('JPATH_BASE') or die;
 
 // JLayout for standard handling of associations fields in the administrator items edit screens.
-if ($displayData->getForm()->getValue('id') != 0 && $displayData->getForm()->getValue('language') != "*")
+if ($displayData->getForm()->getValue('id') != 0 || $displayData->getForm()->getValue('language') != "*")
 {
 	echo $displayData->getForm()->renderFieldset('item_associations');
 }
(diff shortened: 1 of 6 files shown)


layouts/joomla/edit/associations.php
```diff
@@ -10,4 +10,11 @@
 defined('JPATH_BASE') or die;
 
 // JLayout for standard handling of associations fields in the administrator items edit screens.
-echo $displayData->getForm()->renderFieldset('item_associations');
+if ($displayData->getForm()->getValue('id') != 0 && $displayData->getForm()->getValue('language') != "*")
```
andrepereiradasilva, line 13: why not use 

``` php
$displayData->item->id != 0 && $displayData->item->language != '*'
```

like it waS?

jreys, line 13: those fields are protected, it won't work that way, tried it

jreys: it's an OR though, we want that behaviour in either cases

infograf768:
This works fine here.
Now, the fields are present although hidden.

infograf768:
Please propose the patch ASAP towards joomla staging.

infograf768:
@jreys

andrepereiradasilva:
merged in latest joomla staging

PR #27, opened 2016-07-27
Improve again the helper

andrepereiradasilva:
More improvements to the helper so it can be used in other places
#### How to test

All select box option in list view work as before.


jreys:
Tested it, works OK

andrepereiradasilva:
ok thanks. merged

andrepereiradasilva:
as you can ssee now you have the assocuiation helpers file, class, etc
also returns if component supports associations.

i think this helper now does all needed checks (and returns the data) needed.
https://github.com/joomla-projects/gsoc16_improved-multi-lingual/blob/staging/administrator/components/com_associations/helpers/associations.php#L32

if anything more can be improved.

PR #37, opened 2016-07-29
ACL (and checkin) for list and side by side views

jreys:
Pull Request for Issue #7.
#### Summary of Changes

Adding ACL in com_associations
#### Notes
- When a user accesses a forbidden item, it should be redirected back to the list view with a notice/error, I'm aware of that
- Reference is not yet implemented, I'm thinking in using JS to hide the button in case the user is not authorised, it won't compromise security since the save action is triggered by the iframe.


administrator/components/com_associations/views/association/view.html.php
```diff
@@ -147,11 +147,28 @@ protected function addToolbar()
...
+
 		JToolbarHelper::title(JText::_('COM_ASSOCIATIONS_HEADER_EDIT'), 'contract');
 
-		JToolbarHelper::apply('reference', 'COM_ASSOCIATIONS_SAVE_REFERENCE');
+		$canEdit = $user->authorise('core.edit', $component . $this->referenceId);
+		if (isset($this->table->created_by))
+		{
+			$canEditOwn = $user->authorise('core.edit.own', $component . $this->referenceId) && $this->table->created_by == $userId;
+		}
+
```
jreys: Is this the way of getting the authorise? I'm using this since it doesn't need to load a helper like in:
https://github.com/joomla-projects/gsoc16_improved-multi-lingual/blob/staging/administrator/components/com_contact/views/contact/view.html.php#L90

or

https://github.com/joomla-projects/gsoc16_improved-multi-lingual/blob/staging/administrator/components/com_menus/views/item/view.html.php#L57

If this is wrong please tell me @infograf768 @andrepereiradasilva @alikon

administrator/components/com_associations/views/association/view.html.php
```diff
@@ -147,11 +147,28 @@ protected function addToolbar()
 		$input = JFactory::getApplication()->input;
 		$input->set('hidemainmenu', 1);
 
+		$user       = JFactory::getUser();
+		$userId     = $user->id;
+		$checkedOut = !($this->table->checked_out == 0 || $this->table->checked_out == $userId);
```
andrepereiradasilva: @jreys DON'T USE direct table fields!! You can't be ever sure what is the field name of the 3º party component for created_by.

That's why i created this: https://github.com/joomla-projects/gsoc16_improved-multi-lingual/blob/staging/administrator/components/com_associations/helpers/associations.php#L214-L231

Please use it!

andrepereiradasilva: just use
`$this->component->fields->checked_out`

administrator/components/com_associations/views/association/view.html.php
andrepereiradasilva: i'm no expert in this part.. but look at how all other core componets do it and the awnser shown be there

andrepereiradasilva:
more asy to see the changes with https://github.com/joomla-projects/gsoc16_improved-multi-lingual/pull/37/files?w=1

administrator/components/com_associations/views/associations/tmpl/default.php
```diff
@@ -85,6 +85,7 @@
 			<tbody>
 			<?php foreach ($this->items as $i => $item) :
 				$canEdit    = $user->authorise('core.edit', $this->state->get('component') . $item->id);
+				
 				if (isset($item->created_by))
 				{
 					$canEditOwn = $user->authorise('core.edit.own', $this->state->get('component') . $item->id) && $item->created_by == $userId;
```
andrepereiradasilva: replace all $this->state->get('component') by $this->component->component in this ACL rules

why? just make a dump of the vars and you will see.

administrator/components/com_associations/models/fields/menuparent.php
```diff
@@ -0,0 +1,91 @@
...
+ * @package     Joomla.Administrator
+ * @subpackage  com_menus
+ *
+ * @copyright   Copyright (C) 2005 - 2016 Open Source Matters, Inc. All rights reserved.
+ * @license     GNU General Public License version 2 or later; see LICENSE.txt
+ */
+
+defined('JPATH_BASE') or die;
+
+JFormHelper::loadFieldClass('list');
```
andrepereiradasilva: what is this file for??

administrator/components/com_associations/models/fields/associatedcomponent.php
```diff
@@ -64,40 +64,46 @@ protected function getGroups()
...
-			// Only components that exist also in the site client, aren't in the excluded components array and have models.
-			if (!is_dir($componentModelsPath) || in_array($component, $excludeComponents))
-			{
-				continue;
-			}
-
-			// Check if component uses associations, by checking is models.
-			foreach (glob($componentModelsPath . '/*.php', GLOB_NOSORT) as $modelFile)
+			if ($user->authorise('core.manage', $component . $this->referenceId))
 			{
```
andrepereiradasilva: thsi works, but it's not engough

try the following:
- Create a user assigned to "Administrator" group
- Go to com_newsfeeds and say the Administrator groups is Denied in "Access Administration Interface"
- Login with that user in a new private window or other broewser
- Check the components menu news feeds is not there.
- Go to "Multilingual Associations" and the check newsfeed is not in the components list
- All ok until now...
- Now, access /administrator/index.php?option=com_associations&view=associations&component=com_newsfeeds.newsfeed&language=en-GB
 Voliá **ACL breach**...

So you need to make that check also here  (https://github.com/joomla-projects/gsoc16_improved-multi-lingual/blob/staging/administrator/components/com_associations/views/associations/view.html.php#L81) and also you will need to make it in the side by side view.

administrator/components/com_associations/models/fields/menuparent.php
jreys: No idea, but searching:
https://github.com/joomla-projects/gsoc16_improved-multi-lingual/search?utf8=%E2%9C%93&q=JFormFieldList
confirms that it's needed :stuck_out_tongue:

andrepereiradasilva: needed for? if you don't have an idea don't submit a PR with it ....

administrator/components/com_associations/associations.php
```diff
@@ -15,6 +15,17 @@
...
+$splitcpnt = explode('.', $component);
+$component = $splitcpnt[0];
+
+// Check if user has permission to access the component
+if ($component != '' && !JFactory::getUser()->authorise('core.manage', $component))
+{
+    throw new Exception(JText::_('JERROR_ALERTNOAUTHOR'), 403);
+}
+
 JLoader::register('AssociationsHelper', __DIR__ . '/helpers/associations.php');
```
andrepereiradasilva: @jreys  this will not work in all scenarios

something like this is more expected to work. but needs tests:
```php
JLoader::register('AssociationsHelper', __DIR__ . '/helpers/associations.php');

$component = JFactory::getApplication()->input->get('component', '', 'string');

if ($component !== '')
{
	$cp = = AssociationsHelper::getComponentProperties($component);

	if (!$cp[$key]->associations->support)
	{
		throw new Exception(JText::sprintf('The component %s doesn't support associations.', $cp->realcomponent), 404);
	}

	if (!JFactory::getUser()->authorise('core.manage', $cp->realcomponent))
	{
		throw new Exception(JText::_('JERROR_ALERTNOAUTHOR'), 403);
	}
}
```

Also you need to refactor the side by side view URL parameters so this can work there too.

administrator/components/com_associations/views/association/view.html.php
```diff
@@ -147,11 +147,30 @@ protected function addToolbar()
...
+		$user       = JFactory::getUser();
+		$userId     = $user->id;
+		$checkedOut = !($this->table->{$this->component->fields->checked_out} == 0 || $this->table->{$this->component->fields->checked_out} == $userId);
+		$component  = $input->get('acomponent', '', 'string');
+		$createdBy  = $this->component->fields->created_by;
+
 		JToolbarHelper::title(JText::_('COM_ASSOCIATIONS_HEADER_EDIT'), 'contract');
 
-		JToolbarHelper::apply('reference', 'COM_ASSOCIATIONS_SAVE_REFERENCE');
+		$canEdit = $user->authorise('core.edit', $component . "." . $this->associatedView . $this->referenceId);
```
andrepereiradasilva: just reading the code i can see this is not correct ... did you even test?

if should be something like this

```php
$canEdit = $user->authorise('core.edit', $this->component->assetKey . '.' . $this->referenceId);
```

administrator/components/com_associations/views/association/view.html.php
```diff
@@ -147,11 +147,30 @@ protected function addToolbar()
...
+		if (isset($this->table->{$created_by}))
+		{
+			$canEditOwn = $user->authorise('core.edit.own', $component . $this->referenceId) && $this->table->{$created_by} == $userId;
```
andrepereiradasilva: same thing ... if should be something like this

```php
$canEditOwn = $user->authorise('core.edit.own', $this->component->assetKey . '.' . $this->referenceId) && $this->component->{$created_by} == $userId;
```

administrator/components/com_associations/views/associations/tmpl/default.php
andrepereiradasilva: again 
```php
$user->authorise('core.edit', $this->component->assetKey . '.' . $item->id)
[...]
$user->authorise('core.edit.own', $this->component->assetKey . '.' . $item->id)
```

andrepereiradasilva:
joao, i just full reviewed this. Please check my comments,.

Also **please test** properly!

administrator/components/com_associations/models/fields/menuparent.php
jreys: oh I read that wrong, since you commented this line:
`JFormHelper::loadFieldClass('list');`I thought you were referring to it, this file should not be here indeed, I'll remove it

administrator/components/com_associations/views/associations/tmpl/default.php
```diff
@@ -87,11 +87,11 @@
 			</tfoot>
 			<tbody>
 			<?php foreach ($this->items as $i => $item) :
-				$canEdit    = $user->authorise('core.edit', $this->state->get('component') . $item->id);
+				$canEdit    = $user->authorise('core.edit', $this->component->assetKey . $item->id);
```
andrepereiradasilva: you miss the dot separting! as told before.
```php
$this->component->assetKey . '.' . $item->id
```

administrator/components/com_associations/views/associations/tmpl/default.php
```diff
@@ -87,11 +87,11 @@
...
 				
 				if (isset($item->created_by))
 				{
-					$canEditOwn = $user->authorise('core.edit.own', $this->state->get('component') . $item->id) && $item->created_by == $userId;
+					$canEditOwn = $user->authorise('core.edit.own', $this->component->assetKey . $item->id) && $item->created_by == $userId;
```
andrepereiradasilva: you miss the dot separting! as told before.
```php
$this->component->assetKey . '.' . $item->id
```

administrator/components/com_associations/models/fields/itemlanguage.php
```diff
@@ -40,21 +40,22 @@ protected function getOptions()
...
-		$associatedComponent = $input->get('acomponent', '');
-		$associatedView      = $input->get('aview', '');
-		$extension           = $input->get('extension', '');
-		$realView            = $extension !== '' ? $extension : $associatedView;
+		$component           = $input->get('component', '', 'string');
+		$this->component     = AssociationsHelper::getComponentProperties($component);
+		$this->referenceId   = $input->get('id', 0, 'int');
+		$associatedComponent = $this->component->component;
+		$associatedView      = $this->component->item;
+		$realView            = isset($this->component->extension) ? $this->component->extension : $associatedView;
```
andrepereiradasilva: you have $this->component->realcomponent for this ;)

administrator/components/com_associations/views/associations/tmpl/default.php
```diff
@@ -87,10 +87,11 @@
 			</tfoot>
 			<tbody>
 			<?php foreach ($this->items as $i => $item) :
-				$canEdit    = $user->authorise('core.edit', $this->state->get('component') . $item->id);
+				$canEdit    = $user->authorise('core.edit', $this->component->assetKey . $item->id);
+				
 				if (isset($item->created_by))
 				{
-					$canEditOwn = $user->authorise('core.edit.own', $this->state->get('component') . $item->id) && $item->created_by == $userId;
+					$canEditOwn = $user->authorise('core.edit.own', $this->component->assetKey . $item->id) && $item->created_by == $userId;
```
andrepereiradasilva: you miss the dot separating. as told before...
```php
$this->component->assetKey . '.' . $item->id
```

administrator/components/com_associations/models/fields/associatedcomponent.php
```diff
@@ -64,40 +64,46 @@ protected function getGroups()
...
-			// Only components that exist also in the site client, aren't in the excluded components array and have models.
-			if (!is_dir($componentModelsPath) || in_array($component, $excludeComponents))
-			{
-				continue;
-			}
-
-			// Check if component uses associations, by checking is models.
-			foreach (glob($componentModelsPath . '/*.php', GLOB_NOSORT) as $modelFile)
+			if ($user->authorise('core.manage', $component))
 			{
```
andrepereiradasilva: i would put this check later on in the process.
This way it will check ACL for all compoents in the cycle, even some that are excluded in the first place form com_associations.

andrepereiradasilva:
@jreys please fix conflicts and solve the comments

administrator/components/com_associations/associations.php
```diff
@@ -15,6 +15,28 @@
 	return JError::raiseWarning(404, JText::_('JERROR_ALERTNOAUTHOR'));
 }
 
+$input     = JFactory::getApplication()->input;
+$component = $input->get('component', '', 'string');
+$splitcpnt = explode('.', $component);
+$component = $splitcpnt[0];
```
andrepereiradasilva: why this 4 lines above??

administrator/components/com_associations/controllers/association.php
```diff
@@ -27,9 +27,7 @@ class AssociationsControllerAssociation extends JControllerForm
 	 */
 	public function cancel($key = null)
 	{
-		$extension = $this->input->get('extension', '', 'string');
-
-		$key = $extension !== '' ? 'com_categories.category|' . $extension : $this->input->get('acomponent', '') . '.' . $this->input->get('aview', '');
+		$key = $this->input->get('component', '', 'string');
 		$cp  = AssociationsHelper::getComponentProperties($key);
```
andrepereiradasilva: just one line needed
```php
$cp  = AssociationsHelper::getComponentProperties($this->input->get('component', '', 'string'));
```

administrator/components/com_associations/models/fields/associatedcomponent.php
andrepereiradasilva: Maybe instead of
```php
if (!is_dir($componentModelsPath) || in_array($component, $excludeComponents))
```

you do ?
```php
if (!is_dir($componentModelsPath) || in_array($component, $excludeComponents) || !$user->authorise('core.manage', $component))
```

administrator/components/com_associations/associations.php
```diff
@@ -15,6 +15,28 @@
...
+		throw new Exception(JText::_('COM_ASSOCIATIONS_COMPONENT_NOT_SUPPORTED') . " " . $cp->realcomponent, 404);
+	}
+
+	if (!JFactory::getUser()->authorise('core.manage', $cp->realcomponent))
+	{
+		throw new Exception(JText::_('JERROR_ALERTNOAUTHOR'), 403);
+	}
+}
+
 JLoader::register('AssociationsHelper', __DIR__ . '/helpers/associations.php');
```
andrepereiradasilva: remove this line (is already above) ...?

administrator/components/com_associations/models/fields/associatedcomponent.php
```diff
@@ -64,14 +64,17 @@ protected function getGroups()
 			'com_users',
 		);
 
+		$user       = JFactory::getUser();
+		$userId     = $user->id;
```
andrepereiradasilva: where do you use this $userId     = $user->id; ?

administrator/components/com_associations/models/fields/itemlanguage.php
```diff
@@ -40,21 +40,24 @@ protected function getOptions()
 
 		$input = JFactory::getApplication()->input;
 
-		$referenceId         = $input->get('id', '');
-		$associatedComponent = $input->get('acomponent', '');
-		$associatedView      = $input->get('aview', '');
-		$extension           = $input->get('extension', '');
-		$realView            = $extension !== '' ? $extension : $associatedView;
+		$component           = $input->get('component', '', 'string');
+		$this->component     = AssociationsHelper::getComponentProperties($component);
```
andrepereiradasilva: just
```php
$this->component     = AssociationsHelper::getComponentProperties($input->get('component', '', 'string'));
```
and remove line above, not needed

administrator/components/com_associations/models/fields/itemlanguage.php
```diff
@@ -40,21 +40,24 @@ protected function getOptions()
...
+		$this->referenceId   = $input->get('id', 0, 'int');
+		$associatedComponent = $this->component->component;
+		$associatedView      = $this->component->item;
```
andrepereiradasilva: where is this $associatedView used?

administrator/components/com_associations/models/fields/associatedcomponent.php
jreys: nowhere, only needed in core.edit.own

administrator/components/com_associations/models/fields/itemlanguage.php
```diff
@@ -40,21 +40,24 @@ protected function getOptions()
...
 
-		$referenceId         = $input->get('id', '');
-		$associatedComponent = $input->get('acomponent', '');
-		$associatedView      = $input->get('aview', '');
-		$extension           = $input->get('extension', '');
-		$realView            = $extension !== '' ? $extension : $associatedView;
+		$component           = $input->get('component', '', 'string');
+		$this->component     = AssociationsHelper::getComponentProperties($component);
+		$this->referenceId   = $input->get('id', 0, 'int');
+		$associatedComponent = $this->component->component;
```
andrepereiradasilva: why save a var in another var? when yo ucould use just use $this->component->component below.

if ($associatedComponent != 'com_menus')

if ($this->component->component != 'com_menus')

administrator/components/com_associations/views/association/tmpl/edit.php
```diff
@@ -246,9 +246,7 @@
 
 $input      = $this->app->input;
 $layout     = $input->get('layout', '', 'string');
-$aComponent = $input->get('acomponent', '', 'string');
-$aView      = $input->get('aview', '', 'string');
-$extension  = $input->get('extension', '', 'string');
+$component = $input->get('component', '', 'string');
 $rLanguage  = $input->get('referencelanguage', '', 'string') != null ? $input->get('referencelanguage', '', 'string') : '';
```
andrepereiradasilva: this will be alwsys different than null because you set the default value as an empty string

why not just
$rLanguage  = $input->get('referencelanguage', '', 'string');

administrator/components/com_associations/views/association/tmpl/edit.php
```diff
@@ -257,9 +255,8 @@
 </button>
 
 <form action="<?php echo JRoute::_(
-			'index.php?option=com_associations&view=association&layout=' . $layout . '&acomponent='
-			. $aComponent . '&aview=' . $aView . '&extension=' . $extension . '&referencelanguage=' . $rLanguage . '&id='
-			. $this->referenceId
+			'index.php?option=com_associations&view=association&layout=' . $layout . '&component='
+			. $component . $rLanguage . '&id=' . $this->referenceId
```
andrepereiradasilva: this will not work ...

instead of
$component . $rLanguage
it should be
$component . '&referencelanguage=' . $rLanguage

administrator/components/com_associations/views/association/view.html.php
```diff
@@ -62,67 +62,38 @@ public function display($tpl = null)
...
 
 		// Special cases for categories.
-		if ($associatedComponent === 'com_categories')
+		if ($this->component->component === 'com_categories')
 		{
-			$options['view']      = '';
+			$options['view']      = 'category';
 			$options['task']      = 'category.edit';
 			$options['extension'] = $extension;
 		}
```
andrepereiradasilva: you could replace this whole block by just

```php
 		$options = array(
			'option'    => $this->component->component,
  			'view'      => $this->component->item,
  			'task'      => $this->component->item . '.edit',
			'extension' => $this->component->extension,
  			'layout'    => 'edit',
  			'tmpl'      => 'component',
  			'id'        => $this->referenceId,
  		);
```

administrator/components/com_associations/controllers/association.php
```diff
@@ -28,10 +28,8 @@ class AssociationsControllerAssociation extends JControllerForm
 	public function cancel($key = null)
 	{
 		JSession::checkToken() or jexit(JText::_('JINVALID_TOKEN'));
-
-		$extension = $this->input->get('extension', '', 'string');
-
-		$key = $extension !== '' ? 'com_categories.category|' . $extension : $this->input->get('acomponent', '') . '.' . $this->input->get('aview', '');
+		
+		$key = $this->input->get('component', '', 'string');
 		$cp  = AssociationsHelper::getComponentProperties($key);
```
andrepereiradasilva: just
$cp  = AssociationsHelper::getComponentProperties($this->input->get('component', '', 'string'));

administrator/components/com_associations/views/association/view.html.php
```diff
@@ -147,11 +146,31 @@ protected function addToolbar()
...
-		JToolbarHelper::apply('reference', 'COM_ASSOCIATIONS_SAVE_REFERENCE');
+		$canEdit = $user->authorise('core.edit', 'core.edit', $this->component->assetKey . '.' . $this->referenceId);
+		
+		if (isset($this->table->{$created_by}))
+		{
+			$canEditOwn = $user->authorise('core.edit.own', 'core.edit', $this->component->assetKey . '.' . $this->referenceId) && $this->table->{$created_by} == $userId;
+		}
+
+		// ACL for the Save Reference button
+		if (!$checkedOut && ($canEdit || $canEditOwn))
```
andrepereiradasilva: $canEditOwn can be undefined here ... and lead to a php notice

administrator/components/com_associations/views/association/view.html.php
```diff
@@ -147,11 +146,31 @@ protected function addToolbar()
...
+		$createdBy  = $this->component->fields->created_by;
+
 		JToolbarHelper::title(JText::_('COM_ASSOCIATIONS_HEADER_EDIT'), 'contract');
 
-		JToolbarHelper::apply('reference', 'COM_ASSOCIATIONS_SAVE_REFERENCE');
+		$canEdit = $user->authorise('core.edit', 'core.edit', $this->component->assetKey . '.' . $this->referenceId);
+		
+		if (isset($this->table->{$created_by}))
+		{
+			$canEditOwn = $user->authorise('core.edit.own', 'core.edit', $this->component->assetKey . '.' . $this->referenceId) && $this->table->{$created_by} == $userId;
```
andrepereiradasilva: 'core.edit.own' OR 'core.edit' ??

administrator/components/com_associations/views/associations/tmpl/default.php
```diff
@@ -110,7 +111,7 @@
 						<?php if (isset($item->checked_out) && $item->checked_out) : ?>
 							<?php echo JHtml::_('jgrid.checkedout', $i, $item->editor, $item->checked_out_time, 'associations.', $canCheckin); ?>
 						<?php endif; ?>
-						<?php if ($canEdit || $canEditOwn) : ?>
+						<?php if ($canEdit || $canEditOwn && $canCheckin) : ?>
```
andrepereiradasilva: $canEditOwn can be undefined here ... and lead to a php notice

administrator/components/com_associations/views/association/view.html.php
```diff
@@ -132,6 +103,34 @@ public function display($tpl = null)
 		$options['id'] = '';
 		$this->targetLink = 'index.php?' . http_build_query($options);
```
andrepereiradasilva: you could replace this whole block by just

```php
 		$options = array(
			'option'    => $this->component->component,
  			'view'      => $this->component->item,
  			'task'      => $this->component->item . '.edit',
			'extension' => $this->component->extension,
  			'layout'    => 'edit',
  			'tmpl'      => 'component',
  		);

		// Reference item edit link.
		$this->link       = 'index.php?' . http_build_query($options) . '&id=' . $this->referenceId;
		$this->targetLink = 'index.php?' . http_build_query($options) . '&id=0';
```

administrator/components/com_associations/views/association/view.html.php
```diff
@@ -147,11 +146,31 @@ protected function addToolbar()
 		$input = JFactory::getApplication()->input;
 		$input->set('hidemainmenu', 1);
 
+		$user       = JFactory::getUser();
+		$userId     = $user->id;
+		$checkedOut = !($this->table->{$this->component->fields->checked_out} == 0 || $this->table->{$this->component->fields->checked_out} == $userId);
+		$component  = $this->component->component;
```
andrepereiradasilva: $component  = $this->component->component; ... needed for?

administrator/components/com_associations/views/association/view.html.php
```diff
@@ -147,11 +146,31 @@ protected function addToolbar()
...
+		$createdBy  = $this->component->fields->created_by;
```
andrepereiradasilva: $createdBy  = $this->component->fields->created_by; ... needed for?

administrator/components/com_associations/models/fields/itemlanguage.php
```diff
@@ -40,33 +40,32 @@ protected function getOptions()
...
-
-		$referenceLanguage = $table->{$cp->fields->language};
+		$table = clone $this->component->table;
+		$table->load($this->referenceId);
 
 		$existingLanguages = JHtml::_('contentlanguage.existing', false, true);
 
 		foreach ($existingLanguages as $key => $lang)
 		{
-			if ($lang->value == $referenceLanguage)
```
jreys: I was thinking in keeping this for readability, but instead I made a comment :smiley:

administrator/components/com_associations/views/association/view.html.php
```diff
@@ -62,6 +62,36 @@ public function display($tpl = null)
...
+
+		$this->referenceId = $input->get('id', 0, 'int');
+		$this->component   = AssociationsHelper::getComponentProperties($input->get('component', '', 'string'));
+
+		// Get reference language.
+		$this->table = clone $this->component->table;
+		$this->table->load($this->referenceId);
+
+		$this->associatedView = $this->component->item;
+		$extension            = $this->component->extension;
```
andrepereiradasilva: where are this $this->associatedView and $extension used?
can be removed right?

jreys: $this->associatedView is used on the edit.php, removed $extension

andrepereiradasilva: > $this->associatedView is used on the edit.php

are you sure about that? i don't see it anywhere ....

administrator/components/com_associations/views/association/view.html.php
```diff
@@ -62,6 +62,36 @@ public function display($tpl = null)
...
+		$options = array(
+			'option'    => $this->component->component,
+			'view'      => $this->component->item,
+			'task'      => $this->component->item . '.edit',
+			'extension' => $this->component->extension,
+			'layout'    => 'edit',
+			'tmpl'      => 'component',
+		);
+
+		// Reference item edit link.
```
andrepereiradasilva: Should be
// Reference and target item edit links.

infograf768:
Conditions of test:

Create a new User Group as sub of Manager. Create a user for that group.
In com_associations Options, let that User Group Access Administration INterface.

Edit an article and disallow edit, delete edit state to that group. Save the article.

Log with that user with another browser.
Launch com_associations

We get a Notice
Notice: Undefined property: AssociationsViewAssociations::$item in /ROOT/administrator/components/com_associations/views/associations/tmpl/default.php on line 90

I can see the article with its link in the list. Clicking on it loads it in the Side by Side.
In the side by side, The button Save Reference is not present.
I can create a new Reference item, Save Target. Close.

Both articles are checked in and associated.

BUT, if the article is already associated, the associated item does not load in the Target when choosing the language concerned.

administrator/components/com_associations/views/association/view.html.php
jreys: here:
https://github.com/joomla-projects/gsoc16_improved-multi-lingual/pull/37/files/afedfc255de8520858cf5b38802db59202d0bee4#diff-d7a25844a430646eaa1f3071a83a58e0R260

andrepereiradasilva: why not just use $this->component->item there and remove this var?

infograf768:
I also tested the following:
I set the same user group to deny Create a new article in com_contents Options.
When in the side by side view, for any language, the "Category field" is greyed, trying to save gives the error: "Invalid field:  Category"
The Save Target button is present.

Ideal would be to hide the Save Target button, not load the Target form and display a message "You are not authorise to create items in the %s component" where %s is the string for the component name.

administrator/components/com_associations/models/fields/itemlanguage.php
```diff
@@ -40,32 +40,35 @@ protected function getOptions()
 
 		$input = JFactory::getApplication()->input;
 
-		$this->component     = AssociationsHelper::getComponentProperties($input->get('component', '', 'string'));
-		$this->referenceId   = $input->get('id', 0, 'int');
+		$component     = AssociationsHelper::getComponentProperties($input->get('component', '', 'string'));
+		$referenceId   = $input->get('id', 0, 'int');
+		$realView            = isset($component->extension) ? $component->extension : $component->item;
```
andrepereiradasilva: you should use !is_null instead of isset

administrator/components/com_associations/models/fields/itemlanguage.php
andrepereiradasilva: ok

andrepereiradasilva:
### com_associations ACL TEST

#### ACL test for Access Administration Interface (core.manage) in com_associations

##### Menu access to component
- :white_check_mark: Componet menu "Multilingual Associations" does not appear

##### Edit components options: Direct URL acess
- /administrator/index.php?option=com_config&view=component&component=com_associations
  
  :white_check_mark: Denied (error message)

##### List view: Direct URL access
- /administrator/index.php?option=com_associations
  
  :white_check_mark: Denied:  403
- /administrator/index.php?option=com_associations&view=associations
  
  :white_check_mark: Denied:  403
- /administrator/index.php?option=com_associations&view=associations&task=associations.checkin&cid[]=1
  
  :white_check_mark: Denied:  403

#### Side by side view: Direct URL access
- /administrator/index.php?option=com_associations&layout=edit&component=com_content.article&id=3
  
  :white_check_mark: Denied:  403
- /administrator/index.php?option=com_associations&layout=edit&component=com_content.article&id=3&task=association.cancel&target-id=1
  
  :white_check_mark: Denied:  403

#### ACL test for Configure Options Only (core.options) in com_associations

##### Menu access to component
- :white_check_mark: Component menu "Multilingual Associations" appear

##### Edit components options
- /administrator/index.php?option=com_config&view=component&component=com_associations
  
  :white_check_mark: Allowed (cannot edit permissions)

#### ACL test for Configure ACL & Options (core.admin) in com_associations

##### Menu access to component
- :white_check_mark: Component menu "Multilingual Associations" appear

##### Edit components options: Direct URL acess
- /administrator/index.php?option=com_config&view=component&component=com_associations
  
  :white_check_mark: Allowed (can edit permission)

---

So IMO all ok :white_check_mark: for com_associations permisisons.
Good job.
Now its' left to test the componet and component item permissions.

jreys:
Regarding the target item and not yet implemented:
- I need to check if a user can create or edit an item, if can't create, items with id=0 won't show up on the language switcher, if can't edit, items already associated won't show up on the switcher, if can't edit or create, block the view?  my blocking languages from the switcher, won't it be confusing regarding UX?

andrepereiradasilva:
> I need to check if a user can create or edit an item, if can't create, items with id=0 won't show up on the language switcher

Or show up disabled.

> if can't edit, items already associated won't show up on the switcher, 

Or show up disabled.

> if can't edit or create, block the view? my blocking languages from the switcher, won't it be confusing regarding UX?

Or show up disabled.

andrepereiradasilva:
### com_associations component (com_newsfeeds) ACL TEST

#### ACL test for Access Administration Interface (core.manage)

##### Component selector
- :white_check_mark: com_newsfeeds does not appear in the selector, neither his categories

##### List view: Direct URL access
- /administrator/index.php?option=com_associations&view=associations&component=com_newsfeeds.newsfeed
  
  :white_check_mark: Denied:  403
- /administrator/index.php?option=com_associations&view=associations&component=com_newsfeeds.newsfeed&task=associations.checkin&cid[]=1
  
  :white_check_mark: Denied:  403
- /administrator/index.php?option=com_associations&view=associations&component=com_categories.category|com_newsfeeds&task=associations.checkin&cid[]=1
  
  :white_check_mark: Denied:  403

#### Side by side view: Direct URL access
- /administrator/index.php?option=com_associations&layout=edit&component=com_newsfeeds.newsfeed&id=3
  
  :white_check_mark: Denied:  403
- /administrator/index.php?option=com_associations&layout=edit&component=com_newsfeeds.newsfeed&id=3&task=association.cancel&target-id=1
  
  :white_check_mark: Denied:  403
- /administrator/index.php?option=com_associations&layout=edit&component=com_categories.category|com_newsfeeds&id=3&task=association.cancel&target-id=1
  
  :white_check_mark: Denied:  403

#### ACL test for Create (core.create)

##### List view
- click the link to new in the list view
  
  :white_circle: this is still a to do, so no result for now

##### Side by side view
- Select a language without association in the language selector
  
  :white_check_mark: Not allowed. Select option is disabled

#### ACL test for Edit (core.edit)

**(to do)**

#### ACL test for Edit Own (core.edit.own)

**(to do)**

#### Checkin/checkout test

##### List view
- User cannot edit if reference is checkout by another user
  
  :white_check_mark: Not allowed. Error message.
- User cannot edit targets checkout by another user
  
  :white_circle: this is still a to do, so no result for now

##### Side by side view
- User cannot edit if reference is checkout by another user
  
  :white_check_mark: Not allowed. Select option is disabled
- User cannot edit targets checkout by another user
  
  :white_check_mark: Not allowed. Select option is disabled

---

tests with other components (IMO critical: com_menus and com_content)

jreys:
ACL done for target item, will add comments explaining code after lunch

administrator/components/com_associations/models/fields/itemlanguage.php
```diff
@@ -72,17 +74,40 @@ protected function getOptions()
...
 				}
+
+				$canEdit    = $user->authorise('core.edit', $component->assetKey . '.' . $itemId);
+				$table->load($itemId);
+
+				if (!is_null($table->{$component->fields->created_by}))
+				{
+					$canEditOwn = $user->authorise('core.edit.own', $component->assetKey . '.' . $itemId) && $table->{$component->fields->created_by} == $user->id;
+					$canEdit    = $canEdit || $canEditOwn;
+				}
```
andrepereiradasilva: IMHO this could be improved.
you only need to load the item with jtable (extra bd query) if the component supports created_by field, so  ...

```php
if (!is_null($component->fields->created_by))
{
	$table->load($itemId);
	$canEditOwn = $user->authorise('core.edit.own', $component->assetKey . '.' . $itemId) && $table->{$component->fields->created_by} == $user->id;
	$canEdit    = $canEdit || $canEditOwn;
}
```

administrator/components/com_associations/models/fields/itemlanguage.php
```diff
@@ -72,17 +74,40 @@ protected function getOptions()
...
 			}
 			else
 			{
 				$lang->value .= '|0';
+				$canCreate    = $user->authorise('core.create', $component->assetKey);
+
+				if (!$canCreate)
+				{
+					$lang->disable = true;
+				}
```
andrepereiradasilva: additional comment for this cycle.

don't you also need to check the checkout state (if componet support)?

administrator/components/com_associations/models/fields/itemlanguage.php
```diff
@@ -40,47 +40,74 @@ protected function getOptions()
...
```
andrepereiradasilva: Bug! is not $component->assetKey, it is $component->component.
Also can be simplified.
Use just
```php
$lang->value  .= '|0';
$lang->disable = !$user->authorise('core.create', $component->component);
```

administrator/components/com_associations/models/fields/itemlanguage.php
```diff
@@ -40,47 +40,74 @@ protected function getOptions()
...
 			}
+			
 		}
```
andrepereiradasilva: doesn't the checkin needs to be checked here also?
I mean can a user without checkout prividges to a checkedin item be allowed to select it?

administrator/components/com_associations/models/fields/itemlanguage.php
```diff
@@ -40,47 +40,74 @@ protected function getOptions()
...
+				{
+					$canEditOwn = $user->authorise('core.edit.own', $component->assetKey . '.' . $itemId) && $table->{$component->fields->created_by} == $user->id;
+					$canEdit    = $canEdit || $canEditOwn;
+				}
+
+				if (!$canEdit)
+				{
+					$lang->disable = true;
+				}	
 			}
```
andrepereiradasilva: doesn't the checkin needs to be checked here also?
I mean can a user without checkin prividges to a checkin a checkout item (by other user) be allowed to select it in this list?

administrator/components/com_associations/models/fields/itemlanguage.php
andrepereiradasilva: Actually whatt i just said is not correct...

because a category inside the componet can also have core.create persmission.

so we need to check this better

andrepereiradasilva: pinging @infograf768 and @alikon on this

administrator/components/com_associations/models/fields/itemlanguage.php
```diff
@@ -38,49 +38,87 @@ protected function getOptions()
...
+				$canCheckin = !isset($table->{$component->fields->checked_out}) 
+					|| $canManageCheckin 
+					|| $table->{$component->fields->checked_out} == $user->id 
+					|| $table->{$component->fields->checked_out} == 0;
+
+				// If this fails, disable language picking for the user
+				if (!($canEdit && $canCheckin))
+				{
+					$lang->disable = true;
+				}
```
andrepereiradasilva: i would replace this whole edit block by this:

```php
				// Check if user does have permission to edit the associated item.
				$canEdit = $user->authorise('core.edit', $component->assetKey . '.' . $itemId);

				// Check if user does have permission to edit it's own associated item (if component supports it).
				if (!$canEdit && !is_null($component->fields->created_by))
				{
					// Load item.
					$table->load($itemId);

					// Check if user created this item.
					$canEdit = $table->{$component->fields->created_by} == $user->id && $user->authorise('core.edit.own', $component->assetKey . '.' . $itemId);
				}

				// Do an additional check to check if user can edit a checked out item (if component supports it).
				if ($canEdit && !$canManageCheckin && !is_null($component->fields->checked_out))
				{
					// Load item.
					$table->load($itemId);

					// Check if user checked out this item
					$canEdit = in_array($table->{$component->fields->checked_out}, array($user->id, 0));
				}

				// Disable language if user is not allowed to edit the item associated to it.
				$lang->disable = !$canEdit;
```

administrator/components/com_associations/controllers/association.php
```diff
@@ -17,6 +17,40 @@
...
+		if (!$checkedOut && !$cp->model->checkout($id))
+		{
+			$this->setRedirect(JRoute::_('index.php?option=com_associations&view=associations', false));
+			JError::raiseError( 100, JText::sprintf('JLIB_APPLICATION_ERROR_CHECKOUT_FAILED', $cp->model->getError()));
+		}
+		else
+		{
+			return parent::display();
+		}
+	}
```
andrepereiradasilva: improve suggestion
```php
	/**
	 * Method to edit an existing record in side by side view.
	 *
	 * @param   string  $key     The name of the primary key of the URL variable.
	 * @param   string  $urlVar  The name of the URL variable if different from the primary key
	 *                           (sometimes required to avoid router collisions).
	 *
	 * @return  boolean  True if checkout passes, false otherwise.
	 *
	 * @since   __DEPLOY_VERSION__
	 */
	public function edit($key = NULL, $urlVar = NULL)
	{
		$cp = AssociationsHelper::getComponentProperties($this->input->get('component', '', 'string'));

		if (!is_null($cp->fields->checked_out))
		{
			$id    = $this->input->get('id', 0, 'int');
			$user  = JFactory::getUser();
			$table = clone $cp->table;
			$table->load($id);

			// Attempt to check-out the record for editing and redirect.
			if (!in_array($table->{$cp->fields->checked_out}, array($user->id, 0) && !$cp->model->checkout($id))
			{
				JFactory::getApplication()->enqueueMessage(JText::sprintf('JLIB_APPLICATION_ERROR_CHECKOUT_FAILED', $cp->model->getError()), 'error');
				$this->setRedirect(JRoute::_('index.php?option=com_associations&view=associations', false));

				return false;
			}
		}

		return parent::display();
	}
```

administrator/components/com_associations/controllers/association.php
```diff
@@ -17,6 +17,40 @@
...
+	 *
+	 * @param   string  $key     The name of the primary key of the URL variable.
+	 * @param   string  $urlVar  The name of the URL variable if different from the primary key
+	 *                           (sometimes required to avoid router collisions).
+	 *
+	 * @return  boolean  True if access level check and checkout passes, false otherwise.
+	 *
+	 * @since   __DEPLOY_VERSION__
+	 */
+	public function edit($key = NULL, $urlVar = NULL)
```
andrepereiradasilva: the null are lowercase

administrator/components/com_associations/models/fields/itemlanguage.php
```diff
@@ -38,49 +38,90 @@ protected function getOptions()
...
 		// Get reference language.
-		$table = clone $cp->table;
+		$table            = clone $component->table;
 		$table->load($referenceId);
-
-		$referenceLanguage = $table->{$cp->fields->language};
+		$referenceLang    = $table->{$component->fields->language};
+		$user             = JFactory::getUser();
+		$canCreate        = $user->authorise('core.create', $component->realcomponent);
+		$canManageCheckin = $user->authorise('core.manage', 'com_checkin');
```
andrepereiradasilva: as we discovered earlier. remove checkin from here.

administrator/components/com_associations/models/fields/itemlanguage.php
```diff
@@ -38,49 +38,90 @@ protected function getOptions()
...
+				{
+					// Load item.
+					$table->load($itemId);
+
+					// Check if user created this item.
+					$canEdit = $table->{$component->fields->created_by} == $user->id && $user->authorise('core.edit.own', $component->assetKey . '.' . $itemId);
+				}
+
+				// Do an additional check to check if user can edit a checked out item (if component supports it).
+				if ($canEdit && !$canManageCheckin && !is_null($component->fields->checked_out))
```
andrepereiradasilva: as we discovered earlier. remove `&& !$canManageCheckin` checkin from here.

administrator/components/com_associations/views/associations/tmpl/default.php
```diff
@@ -110,7 +113,7 @@
 						<?php if (isset($item->checked_out) && $item->checked_out) : ?>
 							<?php echo JHtml::_('jgrid.checkedout', $i, $item->editor, $item->checked_out_time, 'associations.', $canCheckin); ?>
 						<?php endif; ?>
-						<?php if ($canEdit || $canEditOwn) : ?>
+						<?php if ($canEdit && $canCheckin) : ?>
```
andrepereiradasilva: as we discovered earlier. remove `&& $canCheckin` checkin from here.

administrator/components/com_associations/views/association/view.html.php
```diff
@@ -147,11 +134,32 @@ protected function addToolbar()
 		$input = JFactory::getApplication()->input;
 		$input->set('hidemainmenu', 1);
 
+		$user       = JFactory::getUser();
+		$checkedOut = !($this->table->{$this->component->fields->checked_out} == 0 || $this->table->{$this->component->fields->checked_out} == $user->id);
```
andrepereiradasilva: as we discovered earlier. remove $checkedOut  from here.

administrator/components/com_associations/views/association/view.html.php
```diff
@@ -147,11 +134,32 @@ protected function addToolbar()
...
+		{
+			$canEditOwn = $user->authorise(
+				'core.edit.own', $this->component->assetKey 
+				. '.' . $this->referenceId
+				) && $this->table->{$this->component->fields->created_by} == $user->id;
+			$canEdit    = $canEdit || $canEditOwn;
+		}
+
+		// ACL for the Save Reference button
+		if (!$checkedOut && $canEdit)
```
andrepereiradasilva: as we discovered earlier. remove `!$checkedOut && ` from here.

administrator/components/com_associations/controllers/association.php
```diff
@@ -28,11 +62,8 @@ class AssociationsControllerAssociation extends JControllerForm
 	public function cancel($key = null)
 	{
 		JSession::checkToken() or jexit(JText::_('JINVALID_TOKEN'));
-
-		$extension = $this->input->get('extension', '', 'string');
-
-		$key = $extension !== '' ? 'com_categories.category|' . $extension : $this->input->get('acomponent', '') . '.' . $this->input->get('aview', '');
-		$cp  = AssociationsHelper::getComponentProperties($key);
+
```
infograf768: line 65 should be empty

administrator/components/com_associations/controllers/association.php
```diff
@@ -17,6 +17,40 @@
...
+	 * @param   string  $key     The name of the primary key of the URL variable.
+	 * @param   string  $urlVar  The name of the URL variable if different from the primary key
+	 *                           (sometimes required to avoid router collisions).
+	 *
+	 * @return  boolean  True if access level check and checkout passes, false otherwise.
+	 *
+	 * @since   __DEPLOY_VERSION__
+	 */
+	public function edit($key = null, $urlVar = null)
+	{
```
andrepereiradasilva: you need to check the token here (before any code) to avoid CSRF

```php
JSession::checkToken() or jexit(JText::_('JINVALID_TOKEN'));
```

administrator/components/com_associations/views/association/tmpl/edit.php
```diff
@@ -246,20 +246,17 @@
...
 		data-hide-reference="<?php echo JText::_('COM_ASSOCIATIONS_EDIT_HIDE_REFERENCE'); ?>"><?php echo JText::_('COM_ASSOCIATIONS_EDIT_HIDE_REFERENCE'); ?>
 </button>
 
 <form action="<?php echo JRoute::_(
-			'index.php?option=com_associations&view=association&layout=' . $layout . '&acomponent='
-			. $aComponent . '&aview=' . $aView . '&extension=' . $extension . '&referencelanguage=' . $rLanguage . '&id='
-			. $this->referenceId
+			'index.php?option=com_associations&view=association&layout=' . $layout . '
+			&component=' . $component . '&referencelanguage=' . $rLanguage . '&id=' . $this->referenceId
 		); ?>" method="post" name="adminForm" id="adminForm" class="form-validate" data-associatedview="<?php echo $this->associatedView; ?>">
```
andrepereiradasilva: $this->component->item instead of $this->associatedView
and remove the $this->associatedView var from the view.

administrator/components/com_associations/views/association/view.html.php
```diff
@@ -62,6 +62,35 @@ public function display($tpl = null)
...
+		$input       = $this->app->input;
+
+		$this->referenceId = $input->get('id', 0, 'int');
+		$this->component   = AssociationsHelper::getComponentProperties($input->get('component', '', 'string'));
+
+		// Get reference language.
+		$this->table = clone $this->component->table;
+		$this->table->load($this->referenceId);
+
+		$this->associatedView = $this->component->item;
```
andrepereiradasilva: see comment above

administrator/components/com_associations/models/fields/itemlanguage.php
```diff
@@ -38,49 +38,89 @@ protected function getOptions()
...
+				$lang->disable = !$canEdit;
 			}
 			else
 			{
-				$lang->value .= '|0';
+				// New item, id = 0 and disabled if user is not allowed to create new items
+				$lang->value  .= '|0';
+				$lang->disable = !$canCreate;
 			}
+
```
infograf768: line 123 should be empty

jreys:
Note: Menus not yet fully tested

jreys:
According to my investigation:

https://github.com/joomla-projects/gsoc16_improved-multi-lingual/blob/staging/administrator/components/com_menus/views/items/tmpl/default.php#L116

https://github.com/joomla-projects/gsoc16_improved-multi-lingual/blob/staging/administrator/components/com_menus/controllers/item.php#L59

https://github.com/joomla-projects/gsoc16_improved-multi-lingual/blob/staging/administrator/components/com_menus/controllers/item.php#L192

Menus aren't treated different so it works

andrepereiradasilva:
besides "investigating" did you tested it??

i say this for a simple matter.

menu item (the menu item) typeAlias is `com_menus.item` but menu items DOESN'T have ACL.
See https://github.com/joomla-projects/gsoc16_improved-multi-lingual/blob/staging/administrator/components/com_menus/models/item.php#L30

menu (the menu itself) typeAlias (_context ???) is `com_menus.menu` and menu DOES have ACL.
See https://github.com/joomla-projects/gsoc16_improved-multi-lingual/blob/staging/administrator/components/com_menus/models/menu.php#L34

The question is when you set the ACL permission for a particular menu, IIRC the menu items inherit it, so this component MUST do the same thing.

Try the following:
- Go Menus > Manage (click on the menu type link of a menu) and go to "Menu Permissions" tab
- Disable the "core.edit" for your test user.
- Now login with tthat user
- Go to menus and you'll see that the user cannot edit the menu items for that menu
- Now go to com_associations, select Menu components and voilá **ACL breach**

the same happens to all other menu ACL permissions...

administrator/components/com_associations/controllers/association.php
```diff
@@ -29,9 +31,18 @@ class AssociationsControllerAssociation extends JControllerForm
 	 */
 	public function edit($key = null, $urlVar = null)
 	{
-		JSession::checkToken() or jexit(JText::_('JINVALID_TOKEN'));
+		//JSession::checkToken() or jexit(JText::_('JINVALID_TOKEN'));
```
andrepereiradasilva: of course this can't commented to merge this.

administrator/components/com_associations/controllers/association.php
```diff
@@ -29,9 +31,18 @@ class AssociationsControllerAssociation extends JControllerForm
...
 		$cp = AssociationsHelper::getComponentProperties($this->input->get('component', '', 'string'));
 
+		$table = clone $cp->table;
+		$table->load($this->input->get('id', null, 'int'));
```
andrepereiradasilva, line 37: humm.. what happens when you don't send a id? load(null) ?

yvesh, line 37: @andrepereiradasilva It gives you an jtable instance of the table type without any data, so this is quite valid to useful sometimes.

andrepereiradasilva:
CS issues

```
FILE: ...l/administrator/components/com_associations/controllers/association.php
--------------------------------------------------------------------------------
FOUND 1 ERROR(S) AFFECTING 1 LINE(S)
--------------------------------------------------------------------------------
 34 | ERROR | Please put a space between the // and the start of comment text;
    |       | found "//JSession::checkToken() or
    |       | jexit(JText::_('JINVALID_TOKEN'));"
--------------------------------------------------------------------------------
UPGRADE TO PHP_CODESNIFFER 2.0 TO FIX ERRORS AUTOMATICALLY
--------------------------------------------------------------------------------
FILE: ...gual/administrator/components/com_associations/helpers/associations.php
--------------------------------------------------------------------------------
FOUND 3 ERROR(S) AFFECTING 3 LINE(S)
--------------------------------------------------------------------------------
 268 | ERROR | The comments for parameters $componentKey (1) and $item (2) do
     |       | not align
 292 | ERROR | The comments for parameters $componentKey (1) and $item (2) do
     |       | not align
 342 | ERROR | The comments for parameters $componentKey (1) and $item (2) do
     |       | not align
--------------------------------------------------------------------------------
UPGRADE TO PHP_CODESNIFFER 2.0 TO FIX ERRORS AUTOMATICALLY
--------------------------------------------------------------------------------
```

andrepereiradasilva:
and again i say `JSession::checkToken() or jexit(JText::_('JINVALID_TOKEN'));` cannot be comment out.

jreys:
well... it's commented out because I get the message every time I visit the edit page, can you tell me what's happening? I have no clue

andrepereiradasilva:
well it ... shoud work because you have the token:
- List view: https://github.com/jreys/gsoc16_improved-multi-lingual/blob/acl/administrator/components/com_associations/views/associations/tmpl/default.php#L154
- Side by side view: https://github.com/jreys/gsoc16_improved-multi-lingual/blob/acl/administrator/components/com_associations/views/association/tmpl/edit.php#L292

jreys:
image

everytime I try to access it

andrepereiradasilva:
ok i see you are not POSTing the form, it's a link, so just GETting the url params.
that why you get the warning. because you don't send the token in the edit links.

See how other components do it, if they don't have that check on edit links ... remove it

jreys:
Content:
https://github.com/joomla-projects/gsoc16_improved-multi-lingual/blob/staging/components/com_content/controllers/article.php#L175

Menus:
https://github.com/jreys/gsoc16_improved-multi-lingual/blob/acl/administrator/components/com_menus/controllers/item.php#L192

Contact and Newsfeeds don't override edit
Parent:
https://github.com/joomla-projects/gsoc16_improved-multi-lingual/blob/e15f57f739eed6db9b6e50f22eda21425683df30/libraries/legacy/controller/form.php#L359

So, no need for that

infograf768:
See my comment on chat. We do have core bugs concerning edit_own. GRRR
https://github.com/joomla/joomla-cms/issues/11466

infograf768:
Also, even when it is working Ok, for articles as example:

The user has edit_own for an article but NOT for the associated articles.
If he clicks on the article in the list (com_associations list), the side by side page displays OK, but, as soon as Selecting a language in the Selector (for which we have an association), if the target item was not created by the same user, we get:

screen shot 2016-08-05 at 10 08 22

andrepereiradasilva:
@jreys can you check why core.edit.own is not being check in articles as @infograf768 
is the problem here? https://github.com/joomla-projects/gsoc16_improved-multi-lingual/pull/37/files#diff-0d6d7aa3ac778b0bb560b220330e853aR94

jreys:
I've found yet another problem, I think it's the same:

image

In this case here, pt-PT is new and I have core.create enabled:

image

Nvm, found the problem:

image

Notice that extra '0'

jreys:
Ok fixed that part:

Still, is this a(nother) core bug?

administrator/index.php?option=com_content&view=article&task=article.edit&layout=edit&tmpl=component&id=0

ok, the action here is to create and not to edit, BUT, shouldn't the edit work if id = 0?

infograf768:
will test tomorrow

infograf768:
Looks like we have a conflict with gsoc staging for the file
`administrator/components/com_associations/views/association/tmpl/edit.php`

Other than that, we still have the issue I explained above
https://github.com/joomla-projects/gsoc16_improved-multi-lingual/pull/37#issuecomment-237784481

This imho has to be solved before merging

infograf768:
@andrepereiradasilva @jreys 
The more I think of it, the more I think we should change our way at ACL for com_associations

I.e. we should force a user group to have at minimum Create and Edit permissions for the specific component and we should drop entirely edit.own in com_associations.

When selecting an Item Type for which Create and Edit are not authorized, I see this as non-clickable Items in the list view and getting a Notice (via AJAX)

alikon:
agree, we should block on start... we cannot use a fine granularity at this time...

administrator/components/com_associations/helpers/associations.php
```diff
@@ -260,4 +260,99 @@ public static function getComponentProperties($key = '')
...
+	 *
+	 * @return  boolean.
+	 *
+	 * @since  __DEPLOY_VERSION__
+	 */
+	public static function allowCreate($componentKey = '')
+	{
+		$user = JFactory::getUser();
+		
+		return $user->authorise('core.create', $componentKey->realcomponent);
```
andrepereiradasilva, line 335: @jreys how will this work on com_menus ?

administrator/components/com_associations/helpers/associations.php
```diff
@@ -260,4 +260,99 @@ public static function getComponentProperties($key = '')
...
+	 */
+	public static function allowCheckout($componentKey = '', $item = null)
+	{
+		if (!is_null($componentKey->fields->checked_out))
+		{
+			$user = JFactory::getUser();
+
+			// Check if user checked out this item
+			return in_array($item->{$componentKey->fields->checked_out}, array($user->id, 0));
+		}
```
andrepereiradasilva, line 356: needs `return false;` as fallback.

andrepereiradasilva:
sorry guys i don't agree.
as it is i see ACL as almost implemented.

infograf768:
ACL may be almost OK, but it will NOT work if user has only core.edit.own for a specific component. 
The issue may not be the PR here, but the permissions necessary to work with com_associations.

administrator/components/com_associations/helpers/associations.php
jreys, line 335: Ok, I remember now, if core.create is disabled for menus, user won't be able to create a new item, i.e., when choosing target, it won't be clickable if it's new. Regarding specific permissions for menu parents, if you try to pick a parent, the default ACL from the component itself won't allow you, are you thinking on an extra case where this might break?

andrepereiradasilva:
will merged as meeting today

PR #38, opened 2016-07-29
Fix menu parent id in component view

andrepereiradasilva:
#### Summary of Changes

Access `/administrator/index.php?option=com_menus&view=item&menutype=&layout=edit&tmpl=component`

Before patch Menu parent field shows admin menu items

image

After patch does not show.

### DIFF
--- a/administrator/components/com_menus/models/fields/menuparent.php
+++ b/administrator/components/com_menus/models/fields/menuparent.php
@@ -56,7 +56,7 @@ class JFormFieldMenuParent extends JFormFieldList
 		// Filter by client id.
 		$clientId = $this->getAttribute('clientid');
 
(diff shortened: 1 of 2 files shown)


jreys:
Tested, doesn't break anything I can think off :)

infograf768:
In fact we DO need this in core as we DO have a bug when creating a new menu item in menu items manager.

When a menu has not been set, i.e. when we display the menu item manager with all menu items, and want to create a new menu item. If one clicks on the parent Item dropdown first, it displays all menu items including back-end until a specific menu is chosen.

I suggest to make a PR to core.

infograf768:
@andrepereiradasilva

andrepereiradasilva:
PR done in main CMS

Issue #36, opened 2016-07-28
Bug! joomla cms with impact in com_associations


andrepereiradasilva:
- edit menu association
- as targat select one language that is not associated yet
- Notice the language is selected
- now, select a menu type
- notice the language is now All again which breaks the whole propose!

![untitled screencast 1](https://cloud.githubusercontent.com/assets/9630530/17232110/ce9bd75e-551d-11e6-833a-0d6521c49125.gif)

@infograf768 can you check this? this seems yet another CMS bug

infograf768:
@andrepereiradasilva 
I confirm the issue. It happens also in the normal core Menu Item creation.
Let's say we create a new menu item, set the language, some other parameters and THEN Select the Menu Item Type.

Before selecting the Menu Item Type, we have:
![screen shot 2016-07-29 at 08 33 56](https://cloud.githubusercontent.com/assets/869724/17239860/90c8320c-5567-11e6-9225-67d65fef33de.png)

After selecting, we get
![screen shot 2016-07-29 at 08 34 34](https://cloud.githubusercontent.com/assets/869724/17239874/a9d3c2e8-5567-11e6-8d5c-9ce80cf6da03.png)

Language is not the only field reset. Status and Access are also reset.

infograf768:
We have another error for new menu item in com_associations:
The Parent Item gives:

![screen shot 2016-07-29 at 09 19 43](https://cloud.githubusercontent.com/assets/869724/17240753/a24956cc-556d-11e6-9c4f-f45faac7b460.png)

infograf768:
For this last one, if I add
`$query->where('a.client_id = 0');`
in `JFormFieldMenuParent::getOptions()`
it's already better

infograf768:
and to complete, we need the ajax call in edit.php

here is a diff:

[gsoc_newmenu_parent.diff.zip](https://github.com/joomla-projects/gsoc16_improved-multi-lingual/files/390190/gsoc_newmenu_parent.diff.zip)

jreys:
Tested:
- When selecting menu item type:
  - Language, Status, Access don't reset
  - Layout tab goes missing, don't know if that's intended (forget it, it's specific to certain items)
- Parent item not showing extra stuff

infograf768:
here is a new one where Save works.
I have overriden the menuparent field in com_associations as George told me that it would be better to patch with `$query->where('a.client_id = 0');` for 3.7.

[...]

PR #39, opened 2016-07-29
Fixing language not updating when menu item type is chosen

jreys:
Pull Request for Issue #36 .
#### Testing Instructions

@infograf768 , I removed your ajax since it didn't affect the language, please test if it's still OK or if I missed something :stuck_out_tongue:


infograf768:
No need for the Ajax anymore.

andrepereiradasilva:
i beliebve somehow the real fix should be done in joomla core.
This also happens when creating a menu item normally in joomla cms core.

PR #40, opened 2016-07-29
Fixing associations save

jreys:
#### Summary of Changes

Associations are now saved when the trigger is made instead of frame reload to prevent further problems
#### Testing Instructions

Test saving either new or existing association.


infograf768:
This does not work here for any new Target.
Also I often get the message:
"To define associations, please first save this item and make sure its language is not set to 'All'."
When displaying the empty new target.

Categories are still not checked in.

infograf768:
@jreys 
https://github.com/joomla-projects/gsoc16_improved-multi-lingual/pull/42 works much better except that categories are sill not checked in.

jreys:
@alikon @andrepereiradasilva @infograf768 you can close this, André's patch works better anyways

PR #41, opened 2016-07-29
Improving multiple select UI/UX

jreys:
#### Summary of Changes

This can be a PR made to joomla-cms, I'm just posting here so you can test it.
If this passes your tests, I'll create a JS file or add to an exiting one (you know best) and then import it on every item that has list views.
#### Testing Instructions
- Go to articles, try clicking on a row, it will select the whole row
- Try trashing multiple items by clicking severall rows
- Try the toggle-all button

Don't merge


andrepereiradasilva:
@jreys i will not comment on this now because you have a lot of other more priority things to do.

We can check this out after the project.
So i will close this and them we can reopen this after the project (just don't delete the branch).

PR #42, opened 2016-07-30
Improve js

andrepereiradasilva:
#### Summary of Changes

Refactor js:
- code comments
- use vanilla js where possible
- use joomla api 
- use data-\* attributes
- etc etc

That i tested the only thing that is not totally working after this is the menus because it doesn't uses modals for associations neither have and id field ...!
See https://github.com/joomla-projects/gsoc16_improved-multi-lingual/issues/43

@jreys please test.

### DIFF
--- a/administrator/components/com_associations/models/forms/association.xml
+++ b/administrator/components/com_associations/models/forms/association.xml
@@ -6,8 +6,7 @@
 			type="itemlanguage"
 			label="COM_ASSOCIATIONS_ITEM_FIELD_LANGUAGE_LABEL"
 			description="COM_ASSOCIATIONS_ITEM_FIELD_LANGUAGE_DESC"
-			onchange="loadFrame(jQuery('#jform_itemlanguage').val());">
-			
+			>
 			<option value="">COM_ASSOCIATIONS_SELECT_TARGET_LANGUAGE</option>
 		</field>
 	</fieldset>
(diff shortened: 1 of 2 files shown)


infograf768:
This works much better than #40 
The Categories are sill not checked in.

infograf768:
To solve the checkin for categories, here is a patch:

```
diff --git a/administrator/components/com_associations/helpers/associations.php b/administrator/components/com_associations/helpers/associations.php
index a63f15a..333c3e8 100644
--- a/administrator/components/com_associations/helpers/associations.php
+++ b/administrator/components/com_associations/helpers/associations.php
@@ -97,4 +97,11 @@
                $cp[$key]->associations->gethelper->method = 'getAssociations';
            }
+           // Special for com_categories as helper is in back-end
+           elseif ($cp[$key]->component === 'com_categories')
+           {
+               $cp[$key]->associations->gethelper->class  = 'CategoriesHelper';
+               $cp[$key]->associations->gethelper->file   = $cp[$key]->adminPath . '/helpers/categories.php';
+               $cp[$key]->associations->gethelper->method = 'getAssociations';
+           }

            // If association get items helper class does not exists, component does not support associations.
```

infograf768:
@andrepereiradasilva 
Can you integrate patch above?

PR #44, opened 2016-07-30
Solving categories checkin on Close

infograf768:
As title says. Edit categories side by side. Patch


alikon:
i've tested successfully

jreys:
Tested:
- Checkin works fine for categories

PR #45, opened 2016-07-30
Solving associations problem

jreys:
#### Testing Instructions

Try "playing around" with adding new items with existing associations.

Example:
- Item en-GB has association with it-IT, try creating a new pt-PT


infograf768:
Let's now concentrate on menus

PR #46, opened 2016-07-30
Menu form with id

andrepereiradasilva:
As talked in glip PoC to add the menu id to the menu item edit form

To test: Edit/create a menu item, view html source find jform_id


andrepereiradasilva:
have to merge this or menu side by side will not work

PR #50, opened 2016-07-30
Checkin in list view

andrepereiradasilva:
This PR adds the funcionaly to view the items checkout in list view.

Also it allows to checkin items to any component as long as it accept checkin/checkout

Proof :)
image

@jreys what do you think?

Note: since this adds a column with cjeckboxes will also allow to do more select tasks if needed in the future.


jreys:
I think this component should be renamed, com_centralise_components sounds much better :sunglasses: , the PR looks good

infograf768:
Tested Ok. Merging.

PR #51, opened 2016-07-31
Use php Exceptions instead of deprecated JError 

andrepereiradasilva:
JError is deprecated and will be removed in 4.0
So let's use non deprecated code.
Code review.


alikon:
i've tested successfully

PR #52, opened 2016-07-31
Avoid ordering error if component does not have ordering

andrepereiradasilva:
For testing we would need a component with associations without ordering/lft column.

So, we can simulate that:

After patch add `$cp[$key]->fields->ordering = null;` in this line https://github.com/joomla-projects/gsoc16_improved-multi-lingual/blob/staging/administrator/components/com_associations/helpers/associations.php#L233

And after do a code review.


alikon:
i've tested successfully

PR #53, opened 2016-07-31
Remove hardcoded associations context

andrepereiradasilva:
The associations context is somewhat hardcoded, it should use the associations context defined  in the model, not be hardcoded.
We already have it returned by the helper, so let's use it.
#### Tests
- List view associations column work fine
- Code review.


alikon:
i've tested successfully

PR #54, opened 2016-07-31
Add token check to association cancel method

andrepereiradasilva:
Code review


alikon:
i've tested successfully

PR #48, opened 2016-07-30
Menu list modal layout (for use with custom target buttom)

andrepereiradasilva:
Initial commit to start work on menu items list modal

Use the following URL to test the filters, search, etc

```
Modal:
/administrator/index.php?option=com_menus&view=items&layout=modal&tmpl=component&menutype=

Modal with forced language:
/administrator/index.php?option=com_menus&view=items&layout=modal&tmpl=component&tmpl=component&menutype=&forcedLanguage=en-GB
```

/administrator/index.php?option=com_menus&view=items&menutype=&layout=modal&tmpl=component

PR #57, opened 2016-08-01
Adding modals for menu item editing + assoc

infograf768:
Includes @andrepereiradasilva modified PR https://github.com/joomla-projects/gsoc16_improved-multi-lingual/pull/48 in order to never associate url, separator, heading, alias menu items.
- the necessary files to Select and Edit Associations when editing a menu item in the Menu Item Manager.

NOTE:
Remains to do and help welcome to Save and Close correctly when editing a menu item in its modal. This needs some js somewhere
Also tooltips for the Buttons in the Associations tab.

After patch, edit a menu item on a multingual site, go to Associations tab. You should get:

screen shot 2016-08-01 at 17 05 00


jreys:
Tested OK, also tested on the multilingual side-by-side, works as intended!

See @andrepereiradasilva , this is why I wanted to leave the menu fix for last! :stuck_out_tongue_winking_eye: oh well, at least I learned a bit about chosen fields, will come in handy in the copy button.

infograf768:
Look at the NOtes above: we still miss some stuff.

andrepereiradasilva:
nice works @infograf768 will check when i have time

administrator/language/en-GB/en-GB.com_menus.ini
```diff
@@ -174,6 +176,7 @@ COM_MENUS_OPTION_SELECT_LEVEL="- Select Max Levels -"
 COM_MENUS_REQUEST_FIELDSET_LABEL="Required Settings"
 COM_MENUS_SAVE_SUCCESS="Menu item successfully saved."
 COM_MENUS_SELECT_MENU="- Select Menu -"
+COM_MENUS_SELECT_A_MENUITEM="Select a Menu Item"
```
andrepereiradasilva, line 178: alpha order ... :)

infograf768:
lol

andrepereiradasilva:
:wink:

alikon:
maybe this  illustrate more the previuos note  about pressing the save & close button issue

>  Save and Close correctly when editing a menu item in its modal.

menu_modal_issue

andrepereiradasilva:
@JoomliC since your like the _guru of the modals_ :) can we ask your help on this?

The goal of this PR is to make edit and list view modals for menu items too :) so we can use them in associations.

We need this for this project. 
This PR when ready will be submited to the core.

So, what we would like to ask is if you:
- Can you do a code review
- Can you figure out how to solve what is not working (read comments above)

We would really apreciatte your help on this.

cyrez:
@andrepereiradasilva I'm not a guru, but will be happy to help!
I will test this PR tomorrow. ;-)

cyrez:
About modal close after **Save** and **Save&Close**, just test to add this (not tested, on code review...) : 

```
            if (task !== "item.apply")
            {
                window.parent.jQuery("#menuEdit' . (int) $this->item->id . 'Modal").modal("hide");
            }

```

Here, just after <code>Joomla.submitform</code> : https://github.com/joomla-projects/gsoc16_improved-multi-lingual/pull/57/files#diff-364530e4fe04868db06c7ed1892f3f09R62

Like this:

```
    } else if (task == 'item.cancel' || document.formvalidator.isValid(document.getElementById('item-form')))
    {
        Joomla.submitform(task, document.getElementById('item-form'));

        if (task !== "item.apply")
        {
            window.parent.jQuery("#menuEdit' . (int) $this->item->id . 'Modal").modal("hide");
        }
    }

```

;-)

andrepereiradasilva:
nice! thanks @JoomliC we really apreciate your help!

@infograf768 can you check?

dgrammatiko:
@andrepereiradasilva @infograf768 solution from @JoomliC is working for me. (i was debuting it at the moment 😀)

andrepereiradasilva:
it will be nice to FINNALY have modals for menus! we could even put a button on tinymce for links to menus too :)

administrator/components/com_menus/models/fields/modal/menu.php
```diff
@@ -0,0 +1,252 @@
...
+ * @copyright   Copyright (C) 2005 - 2016 Open Source Matters, Inc. All rights reserved.
+ * @license     GNU General Public License version 2 or later; see LICENSE.txt
+ */
+
+defined('JPATH_BASE') or die;
+
+/**
+ * Supports a modal menu item picker.
+ *
+ * @since  3.7
```
andrepereiradasilva, line 16: better to put `__DEPLOY_VERSION__` in all since tags

dgrammatiko:
@andrepereiradasilva that's the only internal link button that is missing ATM in joomla

administrator/components/com_menus/models/fields/modal/menu.php
```diff
@@ -0,0 +1,252 @@
...
+		JFactory::getLanguage()->load('com_menus', JPATH_ADMINISTRATOR);
+
+		// The active article id field.
+		$value = (int) $this->value > 0 ? (int) $this->value : '';
+
+		// Build the script.
+		$script = array();
+
+		// Select button script
+		$script[] = '	function jSelectMenu_' . $this->id . '(id, title, object) {';
```
yvesh, line 50: You could really save $this->id into a local js variable, having that 20 times inline with php looks not so nice imo :-)

`var id = ' . $this->id ';`

andrepereiradasilva, line 50: or in a html data-\* attribute ;)

yvesh:
@andrepereiradasilva @jreys Could you please consider moving that whole inline JavaScript into a js file? *duck

We have enough of that ugly inline JavaScript already in Joomla. You can then use document.ready / window.load method to set the parameters like id and value.

P.S.: Any reason you are mixing jQuery and plain JS?
P.P.S.: I know it's still working on progress :-) 

Keep up your good work! Thank you!

andrepereiradasilva:
@yvesh this is a copy (with adaptations) of the code of all the other modals...

yvesh:
@andrepereiradasilva yes i still hope for you to make it better, as a good sample - so someone changes them all ;)

cyrez:
@yvesh Don't worry too much about this now, and better to keep the way it is for other modals (even if not the best way, it is already a  global improvement of what was done in 3.5) as i'm working on a new standard script for all modal type form field (article, user, category, newsfeed, contact...).

The base idea is here (look too at @Fedik comments, who gave me the good way to go, with in mind to not use id, and make it working fine in the new repeatable feature of 3.6) : https://github.com/joomla/joomla-cms/pull/10772#issuecomment-225372783

I was too busy last 2 months to work on it, but i've already started to do some task (and promising to simplify all the field layouts!) ;-)
And so, no issue to follow the currently structure and model, as when my PR will be ready, i will update all existed modal fields using the select/edit/clear type, with in bonus a "Add" item directly from the field (so, the PR i'm working on will remove header script by a standard generic script file).

Can't say when i will have finished this, but will have a bit more time to work again on it this month ;-)

infograf768:
@JoomliC 

> About modal close after Save and Save&Close, just test to add this (not tested, on code review...) : 
> 
> ```
> if (task !== "item.apply")
>         {
>             window.parent.jQuery("#menuEdit' . (int) $this->item->id . 'Modal").modal("hide");
>         }
> ```

That is the first thing I did... But we have a problem with the quotes there (we have `$script = "` and I always get an error. Can you test and propose a patch?

screen shot 2016-08-04 at 09 50 11

@dgt41
I have no idea how you could get this to work...

( ! ) Parse error: syntax error, unexpected 'item' (T_STRING) in ROOT/administrator/components/com_menus/views/item/tmpl/edit.php on line 64

I tried every trick I know of, including using \
for example changing double quotes to single quotes but we lose the `$this->item->id`

```
        if (task !== 'item.apply')
        {
            window.parent.jQuery('#menuEdit' . (int) $this->item->id . 'Modal').modal('hide');
        }
```

I always get:
Catchable fatal error: Object of class stdClass could not be converted to string in ROOT/administrator/components/com_menus/views/item/tmpl/edit.php on line 66

jreys:
@infograf768 , try the following:

```
  if (task !== 'item.apply')
  {
      window.parent.jQuery('#menuEdit" . (int) $this->item->id .  "Modal').modal('hide');
  }
```

edit: removed single quotes after #menuEdit and before Modal
2nd edit: removed spaces

infograf768:
If I do that, I indeed do not get errors anymore when editing a menu item, but I lose the Tabs...
screen shot 2016-08-04 at 10 26 08

infograf768:
yeah, better without the single quotes... Now testing modal

infograf768:
Ok, now remains to solve the Tips on the buttons, as we have now
tips

infograf768:
Once this is done, I think I can propose it to core (yes, with`_DEPLOY_VERSION_` :)

infograf768:
OK, I corrected the tooltip display in the Associations tab for the buttons.

Discovered that we have a lot of discrepancies in core concerning tooltip display in modals. But that will be for another time.
We should now get
screen shot 2016-08-04 at 12 36 57

infograf768:
If all agree, I think we can propose this to Core.

cyrez:
@infograf768 yes, single / double quotes depends on how the script declaration is opened (mainly opened by a single quote, and use of double quote when value/id, but not a big deal, it's ok like this now!) ;-)

> Discovered that we have a lot of discrepancies in core concerning tooltip display in modals. But that will be for another time.

Don't know exactly what you mean, but some differences are due to the fact content in modal is not always the same : layout or iframe (iframe is what is about here)
In the same time, some differences are just ways to fix Bootstrap 2 and 3 tooltip inside modals issues (which were never solved in Bootstrap itself).

About the need to load Tooltip script in the modal field here is because the menu edit view does not set tooltip (but article, category... edit view do!).

That's why the difference ;-)

And if we need to set the bootstrap tooltip is mainly for a B/C reason : bootstrap use "tooltip" class, and Joomla historically use "hasTooltip" from mootools time ;-)
(a think we will be able, and have to clean in J4 ;-) )

infograf768:
@JoomliC 
I had checked what calls we had in edit.php and I did not see any difference concerning setting tooltip in com_menus vs com_contents. What is the exact code you are referring to?

andrepereiradasilva:
IMHO this PR should be moved to joomla cms repository and continue discussions there.
can you do it @infograf768 ?

infograf768:
Yep, I will do it as a new PR there.

infograf768:
Done. 
See https://github.com/joomla/joomla-cms/pull/11471

infograf768:
I suggest to keep this one here as it will be handy to test menu items associations.

andrepereiradasilva:
ok

andrepereiradasilva:
we don't need this for testing now.
i'm gone close.

PR #58, opened 2016-08-02
Hiding target button instead of reference, fixing old PR

jreys:
#### Summary of Changes

During ACL PR this bug is found and this is the fix
#### Testing Instructions
- Check if target button is correctly hidden
- Check if the button styling is still good (I can't have 2 apply buttons, otherwise I will have the same html ID on 2 elements, check article edit view, I did similar)


andrepereiradasilva:
tested ok. simple merge

PR #59, opened 2016-08-03
Undo associations button

jreys:
#### Summary of Changes

Undo associations button added, it has the following behaviour:
- Removes target association from reference
- Removes all associations in target
- Saves reference and target
- Resets language switcher
#### Testing Instructions
- Select item from the list-view (with associations)
- Go to language that has association
- Click button
- Either pick the same language to see if it's empty or close view to check on the list view if it's gone

Note: Menu's not tested yet, will test after lunch


andrepereiradasilva:
A nice idea (for future) would be to add a "Delete associations" on list view.
There you would select a group os items (using thew checkboxes) and just undo all associations for those items.
This, of cours ia a future idea. We probably don't have time to implement in this project.

administrator/components/com_associations/language/en-GB/en-GB.com_associations.ini
```diff
@@ -30,4 +30,5 @@ COM_ASSOCIATIONS_SAVE_TARGET="Save Target"
 COM_ASSOCIATIONS_SELECT_TARGET_LANGUAGE="- Select Target Language -"
 COM_ASSOCIATIONS_SELECT_MENU="- Select Menu -"
 COM_ASSOCIATIONS_TITLE="Associations"
+COM_ASSOCIATIONS_UNDO_ASSOCIATION="Undo Association"
```
andrepereiradasilva: Shouldn't it be called "Remove Association"?

administrator/components/com_associations/views/association/tmpl/edit.php
```diff
@@ -36,6 +37,63 @@
...
+
+			// Save both items
+			Joomla.submitbutton('reference');
+			Joomla.submitbutton('target');
+
+			currentSwitcher = $('#jform_itemlanguage').val();
+			currentLang = referenceLang.replace(/_/,'-');
+			$('#jform_itemlanguage option[value=\"' + currentSwitcher + '\"]').val(currentLang + '|0');
+			$('#jform_itemlanguage').val('').change().trigger('liszt:updated');
+
```
andrepereiradasilva: remove extra line space

andrepereiradasilva:
side comment: after this PR we clearly need to make a PR to move all js to a separated file.

It's already to big to be inline js, and we still have the copy butotn to do ...

jreys:
Tested menus now:
- Opened from list-view
- Removed association on a language
- Selected that language again (item gone)
- Created a new menu-item in that language
- Saved
- Removed association

Worked :smiley:

jreys:
@andrepereiradasilva , I agree in moving the JS to a file, where do you suggest I place it?

andrepereiradasilva:
/media/com_associations/js/sidebyside.js

look at the others /media/com_xxxxxx directories

infograf768:
This works fine here BUT I think we should have ALSO the possibility of Changing an existing association to another existing item AND Select a new association among existing items for a specific language when there are none.

administrator/components/com_associations/views/association/tmpl/edit.php
```diff
@@ -36,6 +37,62 @@
...
+			target.find('#jform_associations_' + referenceLang + '_id').val('');
+			target.find('#jform_associations_' + referenceLang + '_name').val('');
+
+			// - For chosen association selectors (menus).
+			target.find('#jform_associations_' + referenceLang + '_chzn').remove();
+			chznField = target.find('#jform_associations_' + referenceLang).val('').change().chosen();
+
+			// Save both items
+			Joomla.submitbutton('reference');
+			Joomla.submitbutton('target');
```
andrepereiradasilva, line 89: is it really need to submit both forms??

jreys:
Closed, different approach chosen

PR #60, opened 2016-08-05
Move js

jreys:
#### Summary of Changes

As mentioned in #59 , the JS was moved elsewhere so things are more organised.


administrator/components/com_associations/views/association/tmpl/edit.php
```diff
@@ -13,202 +13,7 @@
...
-				}
-			}
-
-			// Iframe load finished, hide Joomla loading layer.
-			Joomla.loadingLayer('hide');
-		}
-	});
-});
-");
+$this->app->getDocument()->addScript(JURI::root( true ) . '/media/com_associations/js/sidebyside.js');
```
andrepereiradasilva: please use
JHtml::_('script', 'com_associations/sidebyside.js', false, true);

andrepereiradasilva:
tested ok. Will be merged.

PR #63, opened 2016-08-06
ACL improve

andrepereiradasilva:
Problems i notice: See https://github.com/joomla/joomla-cms/issues/11466
#### com_contact contacts
- Setting the contact component `core.edit` to `No` and `core.edit.own` to `Yes` should allow the user to edit the items he created inside any category, but core does not allows.
- Setting a contact category `core.edit` to `No` and `core.edit.own` to `Yes` should allow the user to edit the items he created inside that category, but core does not allows.
#### com_newsfeeds newsfeeds
- Setting the newsfeeds component `core.edit` to `No` and `core.edit.own` to `Yes` should allow the user to edit the items he created inside any category, but core does not allows.
- Setting a newsfeed category `core.edit` to `No` and `core.edit.own` to `Yes` should allow the user to edit the items he created inside that category, but core does not allows.

com_associations allows it. So it's a bug that needs to be solved in the core.

The rest seems to be working.
Please test.

### DIFF
--- a/administrator/components/com_associations/views/associations/tmpl/default.php
+++ b/administrator/components/com_associations/views/associations/tmpl/default.php
@@ -18,7 +18,7 @@ JHtml::_('formbehavior.chosen', 'select');
 
 $listOrder        = $this->escape($this->state->get('list.ordering'));
 $listDirn         = $this->escape($this->state->get('list.direction'));
-$canManageCheckin = $user->authorise('core.manage', 'com_checkin');
+$canManageCheckin = JFactory::getUser()->authorise('core.manage', 'com_checkin');
 $colSpan          =  5;
 $iconStates       = array(
 	-2 => 'icon-trash',
--- a/administrator/components/com_associations/views/associations/tmpl/default.php
+++ b/administrator/components/com_associations/views/associations/tmpl/default.php
@@ -9,6 +9,8 @@
 
 defined('_JEXEC') or die;
 
+JLoader::register('AssociationsHelper', JPATH_ADMINISTRATOR . '/components/com_associations/helpers/associations.php');
+
 JHtml::_('jquery.framework');
 JHtml::_('bootstrap.tooltip');
 JHtml::_('behavior.multiselect');
@@ -86,7 +88,7 @@ $iconStates = array(
 			<tbody>
 			<?php foreach ($this->items as $i => $item) :
 				$canEdit    = AssociationsHelper::allowEdit($this->component, $item);
-				$canCheckin = AssociationsHelper::allowCheckout($this->component, $item);
+				$canCheckin = AssociationsHelper::allowCheckActions($this->component, $item);
 				?>
 				<tr class="row<?php echo $i % 2; ?>">
 					<td class="center">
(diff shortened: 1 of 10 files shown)


infograf768:
@andrepereiradasilva 
Did not test this yet, but before this PR, we can't click on a **category** in the list view for which user has only core.edit.own, whatever the component.

administrator/components/com_associations/views/associations/tmpl/default.php
```diff
@@ -9,15 +9,18 @@
...
 JHtml::_('behavior.multiselect');
 JHtml::_('formbehavior.chosen', 'select');
 
-$listOrder  = $this->escape($this->state->get('list.ordering'));
-$listDirn   = $this->escape($this->state->get('list.direction'));
-$colSpan    =  5;
-$iconStates = array(
+$listOrder        = $this->escape($this->state->get('list.ordering'));
+$listDirn         = $this->escape($this->state->get('list.direction'));
+$canManageCheckin = $user->authorise('core.manage', 'com_checkin');
```
infograf768: Missing $user       = JFactory::getUser();

infograf768:
Same issue with core.edit.own (after correcting the comment above)

screen shot 2016-08-07 at 09 00 01

andrepereiradasilva:
> Did not test this yet, but before this PR, we can't click on a category in the list view for which user has only core.edit.own, whatever the component.

IIRC that is also solved by this PR. The categories asset keys has incorrect in the helper.
See https://github.com/joomla-projects/gsoc16_improved-multi-lingual/pull/63/files#diff-b28d30bac596b47de9a66a77481ec94eL203

> Missing $user = JFactory::getUser();

corected. sorry.

> Same issue with core.edit.own (after correcting the comment above)

The issue is not from this PR, if rom the core, test with https://github.com/joomla/joomla-cms/pull/11502 also applied.

infograf768:
> The issue is not from this PR, if rom the core, test with joomla/joomla-cms#11502 also applied.

This is WITH both contacts and newsfeeds PRs applied...

As for categories, it is solved in this PR but same issue with core.edit.own

andrepereiradasilva:
sorry JM, can you do step by step how you got that error, so i can fix it?

infograf768:
As usual
Permissions
screen shot 2016-08-07 at 10 40 14
Create a contact in manager (same for Newsfeeds)
save
Go to list view and click on that contact
You get the side by side view attached above

andrepereiradasilva:
This is how i reproduce that problem:
- Applied newsfeed core.edit.own patch
- Setting com_newsfeeds component `core.edit` to `No` and `core.edit.own` to `Yes` to "Administrator" group
- Created a newsfeed with the ser in "Administrator"
- Edited that newsfeed in side by side
- the whent to the language selector and selected a language and the error appeared

so the problem i see is in the create action on language select

infograf768:
same, except I used a specific group :)

andrepereiradasilva:
ok i see the problem.

the action should be `newsfeed.add` and not `newsfeed.edit` when creating a new one.

This is valid for all componets

andrepereiradasilva:
merging as the previous bug is not and ACL bug

PR #64, opened 2016-08-07
Solve create problem

andrepereiradasilva:
#### Summary of Changes

the create action task is not correct on language select
this Pr solves it
#### Testing Instructions

Clear browser cache (sidebyside.js)
- Apply newsfeed core.edit.own patch
- Setting com_newsfeeds component `core.edit` to `No` and `core.edit.own` to `Yes` to "Administrator" group
- Created a newsfeed with the ser in "Administrator"
- Edited that newsfeed in side by side
- the whent to the language selector and selected a language and the error appeared

@infograf768 

PS: blaim @jreys :wink:


infograf768:
This solves the previous issue

alikon:
even with some delay tested ok

PR #69, opened 2016-08-07
List view assoc links PoC

andrepereiradasilva:
List view direct association edit links.

image

Features:
- Direct links to edit and create.
- Links itself only show if user has ACL permissions.

Note: cs still to do.


infograf768:
IMHO, this should remain as POC. It is not in the scope of the project. Better concentrate on the remaining issues. 😄

andrepereiradasilva:
@infograf768 IMO this is in the scope of the project and, as per timeline, should already be done.
See https://github.com/joomla-projects/gsoc16_improved-multi-lingual/wiki#week-5-20160731---20160806

> Create a new item (article, contact, etc.) by changing language or directly on list view
> Create a new section (category, etc.) by changing language or directly on list view

Also we had a lot of discussion in the chat about this (remember @alikon)
Also i tested this and was working fine, before recent merge.

I just need to fix conflicts.

andrepereiradasilva:
i'm going to close this as it seems the branch got messed up

PR #71, opened 2016-08-08
Flexible db fields (PoC)

andrepereiradasilva:
Flexible db fields.

Imporoves:
- Avoids one db query for component to get the table fields.
#### Test

All items in com_association list views work as before.
#### 3º Party Support

If 3º party use the same fields as JTable doesn't to do anything.
If 3º party does not use the same fields as JTable, needs to add mapping property to his JTable overrides.


andrepereiradasilva:
need to fix conflicts ans use the getColumnsAlias function

andrepereiradasilva:
closed as this was repalced

PR #79, opened 2016-08-09
Removing associations tab buttons

### LINKED ISSUE #76: Show associations on reference
Associations should display on the reference, with no buttons and when target changes, auto-switch to general tab so no warning.

jreys:
Pull Request for Issue #76 .
#### Summary of Changes

See commits
#### Testing Instructions
- Check if associations tab on reference has no buttons and correct text
- Check if associations show on target

If this passes close #76


infograf768:
This works fine here.

PR #80, opened 2016-08-09
Flexible db fields (redo)

andrepereiradasilva:
Flexible db fields.

Improves:
#### Test

All items in com_associations list views work as before.
#### 3º Party Support

If 3º party use the same fields as defined doesn't to do anything.
If 3º party does not use the same fields, needs to add `_columnAlias` property to his JTable override.


andrepereiradasilva:
one question i have, is this the correct order to do the column alias?

``` php
    protected $_columnAlias = array(
        'title'     => 'name',
        'published' => 'state',
    );
```

or should be the opposite?

``` php
    protected $_columnAlias = array(
        'name'  => 'title',
        'state' => 'published',
    );
```

andrepereiradasilva:
messed up again. closing this

Issue #70, opened 2016-08-08
Testing 3rd party component sermonspeaker


infograf768:
I first installed it and got 2 errors:
[08-Aug-2016 07:54:48 UTC] PHP Notice: Undefined index: associations in /Applications/MAMP/htdocs/gsocmulti/administrator/components/com_categories/models/category.php on line 575 
[08-Aug-2016 07:54:48 UTC] PHP Warning: constant(): Couldn't find constant JPATH_ in /Applications/MAMP/htdocs/gsocmulti/libraries/cms/installer/adapter/module.php on line 209

I then tested Associations list:
I only could select Categories and had an unstranslated string:
`COM_SERMONSPEAKER_CATEGORIES`

=> This means imho that we should, when we have categories for a component, force the string to be
`COM_CATEGORIES="Categories"` and not rely on the component specific string as it may not exist.
Change should be line 103 of `ROOT/administrator/components/com_associations/models/fields/associatedcomponent.php`
I.e. chnage
`$languageKey           = JText::_(strtoupper($cp->realcomponent) . '_CATEGORIES');`
to
`$languageKey           = JText::_('COM_CATEGORIES');`

=>Although the component has an association site helper including various views, I did not get them in the dropdown...

infograf768:
I could create and associate categories

jreys:
hmmmm, that sounds a little "hacky", I think we should load the component's language strings, in fact it's being done in
`ROOT/administrator/components/com_associations/helpers/associations.php`
(line 177)
but for some reason, it's not being loaded

infograf768:
The component strings ARE loaded.
But he does not have that string in its ini file...

jreys:
I see now, then I agree, if lang string is not present, we use COM_CATEGORIES instead, this keeps flexibility in case an extension wants to give another name to it

infograf768:
We have a specific code to check if a lang string is present or not
looking now

infograf768:
if ($lang->hasKey( etc.

infograf768:
i.e.

```
            // Check if component uses categories with associations. Add category option to select box if so.
            if ($cp->enabled && $cp->associations->supportCategories)
            {
                $lang = JFactory::getLanguage();

                // Check if component ini file has a specific key
                if ($lang->hasKey($cp->realcomponent . '_CATEGORIES'))
                {
                    $languageKey = JText::_(strtoupper($cp->realcomponent) . '_CATEGORIES');
[...]

PR #84, opened 2016-08-10
Solving tmpl=component issue

jreys:
Pull Request for [this comment](https://github.com/joomla-projects/gsoc16_improved-multi-lingual/issues/70#issuecomment-238791019)

### DIFF
--- a/media/com_associations/js/sidebyside.js
+++ b/media/com_associations/js/sidebyside.js
@@ -150,6 +150,12 @@ jQuery(document).ready(function($) {
 
 	// Attach behaviour to reference frame load event.
 	$('#reference-association').on('load', function() {
+		// If no tmpl=component, add it
+		if (this.src.indexOf('&tmpl=component') === -1)
+		{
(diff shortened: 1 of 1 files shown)


jreys:
This might not be the best approach since the load is being done twice, can't find another way though

infograf768:
will test tomorrow

infograf768:
@jreys This does not work

infograf768:
If I comment this way

``` js
        //if (this.src.indexOf('&tmpl=component') === -1)
        //{
            this.src += '&tmpl=component';
        //}
```

Then it works after loading twice indeed. BUt it gets into a loop.

andrepereiradasilva:
IMHO, shouldn't be solved by com_associations
3º party component need to correctly support tpml=component to work in com_associations

we should drop this PR and made that clear in the documentation

alikon:
> we should drop this PR and made that clear in the documentation

should be a good solution

infograf768:
Closed then...

PR #86, opened 2016-08-12
Proposal: add Clearing the associations table

infograf768:
In some cases, one would like to clear all the records from the associations table.
This PR adds a Toolbar button in the list view to do so.
The button will display only when no Item Type is selected. It displays an alert when clicked.

screen shot 2016-08-12 at 09 43 22

screen shot 2016-08-12 at 09 44 09

screen shot 2016-08-12 at 09 44 44


administrator/components/com_associations/language/en-GB/en-GB.com_associations.ini
```diff
@@ -29,6 +29,11 @@ COM_ASSOCIATIONS_HEADING_MENUTYPE_ASC="Menu ascending"
 COM_ASSOCIATIONS_HEADING_MENUTYPE_DESC="Menu descending"
 COM_ASSOCIATIONS_NO_ASSOCIATION="There is no association for this language"
 COM_ASSOCIATIONS_NOTICE_NO_SELECTORS="Please select an Item Type and a reference language to view the associations."
+COM_ASSOCIATIONS_PURGE="Clear all associations"
+COM_ASSOCIATION_PURGE_CONFIRM_PROMPT="Are you sure you want to clear all associations? Confirming will permanently delete them!"
+; Do not translate the name of the table
+COM_ASSOCIATIONS_PURGE_FAILED="Failed to clear the 'associations' table."
+COM_ASSOCIATIONS_PURGE_SUCCESS="The 'associations' table was successfully cleared."
```
andrepereiradasilva: Proposal:
COM_ASSOCIATIONS_PURGE="Delete all associations"
COM_ASSOCIATION_PURGE_CONFIRM_PROMPT="Are you sure you want to delete all associations? Confirming will permanently delete them!"
COM_ASSOCIATIONS_PURGE_FAILED="Failed to delete all the associations."
COM_ASSOCIATIONS_PURGE_SUCCESS="All associations have been deleted."

infograf768:
I am now working also on a Rebuild table button to get rid of possible orphans.

administrator/components/com_associations/language/en-GB/en-GB.com_associations.ini
infograf768: Core uses "clear"

administrator/components/com_associations/models/associations.php
```diff
@@ -348,4 +348,35 @@ protected function getListQuery()
 
 		return $query;
 	}
+
+	/**
+	 * Cleans out _associations table.
+	 *
+	 * @return  bool True on success
+	 *
+	 * @throws	Exception
```
andrepereiradasilva: it doesn't trow an exception.

administrator/components/com_associations/models/associations.php
```diff
@@ -348,4 +348,35 @@ protected function getListQuery()
...
+	 * @since   __DEPLOY_VERSION__
+	 */
+	public function purge()
+	{
+		$db = $this->getDbo();
+
+		// Get the localise data
+		$db->setQuery('TRUNCATE TABLE #__associations');
```
andrepereiradasilva: Proposal

``` php
// Delete all table data.
$db->setQuery('TRUNCATE TABLE ' . $db->quoteName('#__associations'));
```

administrator/components/com_associations/models/associations.php
```diff
@@ -348,4 +348,35 @@ protected function getListQuery()
...
+
+		try
+		{
+			$db->execute();
+		}
+		catch (JDatabaseExceptionExecuting $e)
+		{
+			$this->_message = JText::_('COM_ASSOCIATIONS_PURGE_FAILED');
```
andrepereiradasilva:
``` php
JFactory::getApplication()->enqueueMessage(JText::_('COM_ASSOCIATIONS_PURGE_FAILED', 'error'));
```

administrator/components/com_associations/language/en-GB/en-GB.com_associations.ini
andrepereiradasilva: image

infograf768: That is because these are files.

Look rather at:
screen shot 2016-08-12 at 10 38 53

and for finder Index
screen shot 2016-08-12 at 10 40 41

infograf768: I do agree though that it is not consequent in core

screen shot 2016-08-12 at 10 43 44

andrepereiradasilva: the core yet again does not make sense
the trash icon is for trashing ie changing it's state to trash in the db
the delete icon is for deleting ie deleting from db

andrepereiradasilva: and BTW the update will be changed in https://github.com/joomla/joomla-cms/pull/11423
so the only bad one is the finder

alikon:
I like it

alikon:
i like this feature

alikon:
i've tested successfully 

RFC
in this way we can only "delete all" associated items
what about being able to  be more granular  i.e. delete all:
- associated articles
-  associated articles categories
-  associated contacts
-  associated contacts catgegories
-  etc

p.s.
tested before the addition of orphans deleting

infograf768:
Corrected lang strings (to please @andrepereiradasilva  😄 )

Implemented cleaning orphans

screen shot 2016-08-13 at 08 15 02

To test orphans deleting, associate 2 items.
With PhpMyAdmin, delete one of them in the associations table so that an orphan remains
For example you would get here an orphaned menu item association:

screen shot 2016-08-12 at 11 33 40

Click on the button.

infograf768:
Concerning more granular delete.
Not so easy as the associations table does not differentiate for example between categories, languages, etc., only context which corresponds to $associationsContext
Therefore much more complex to code

alikon:
we should change the icon of one of the 2 buttons they are the same , maybe can be confusing

infograf768:
btw, I am conscious that the clean Orphans method can be optimized by using simplified queries as I use 3. @alikon

infograf768:
For the Icon, we could use for one of them the Empty Trash icon (i.e. delete)

administrator/components/com_associations/views/associations/view.html.php
```diff
@@ -181,6 +181,11 @@ protected function addToolbar()
 
 		if ($user->authorise('core.admin', 'com_associations') || $user->authorise('core.options', 'com_associations'))
 		{
+			if (!isset($this->component))
+			{
+				JToolbarHelper::custom('associations.purge', 'purge', 'purge', 'COM_ASSOCIATIONS_PURGE', false, false);
+				JToolbarHelper::custom('associations.clean', 'purge', 'purge', 'COM_ASSOCIATIONS_DELETE_ORPHANS', false, false);
```
alikon: can we use refresh instead of purge for orphans button

alikon:
@infograf768  not so sure if we can simplify using 1 complex query instead of the 3 as it is now, but i'll look at it

alikon:
testing the orphans feature:
if i don't have orphans  should tell me something like nothing to delete/no orphans found
instead of the standard message
orphans

infograf768:
> if i don't have orphans should tell me something like nothing to delete/no orphans found
> instead of the standard message

Good idea.

infograf768:
@alikon 
Thanks for the simplified query.

Changed icon, added message when no orphans to delete.
screen shot 2016-08-13 at 10 17 52

Please test again.

alikon:
ok  for me now
just still some concern about delete orphans performance 
if  we need a no-unique index  on key field ..... i need further investigation on this 

and for a proper message if there is nothing to de-associate/clear all associations

but these apart,,,:thumbsup:

infograf768:
Added message when Deleting all associations and the table was empty:
`COM_ASSOCIATIONS_PURGE_NONE="There were no associations to delete."`

andrepereiradasilva:
merging as decide. will add ACL checks in the controller later

PR #88, opened 2016-08-12
Remove htmlhelper - it's unneeded

andrepereiradasilva:
Remove the html helper since now is not needed.
#### Test
1. Apply patch
2. All works fine


andrepereiradasilva:
closing

PR #89, opened 2016-08-12
Do not depend on the helper existing

andrepereiradasilva:
This PR improves the helpers so com_associations doesn't need to depend on the association helper exist.

This avoids errors and improves 3º party support.

Also improves the code.
#### Test
1. Apply patch
2. Item language selector works as before


andrepereiradasilva:
closing

PR #90, opened 2016-08-12
Refactor Helper & component to itemType

andrepereiradasilva:
#### Summary of Changes

This is a refactor of the component.
1. Stop calling component when we are using the item type. All methods, vars, field, etc were changed to reflex this.
2. Now the component select box uses the typeAlias (ex: com_content.category) instead of the earlier costum made string (ex: com_categories.category:com_content)
3. The helper got major refactor to support multiple category extensions
4. More logic and detachable code in the helper.
5. Adds com_associations to the list of components not to search
#### Testing Instructions
1. Apply patch (The patchtester doesn't delete/create new files so you also need to copy the /models/fields/itemtype.php manually - https://raw.githubusercontent.com/joomla-projects/gsoc16_improved-multi-lingual/refactor-helper/administrator/components/com_associations/models/fields/itemtype.php)
2. Logout to clear session data
3. Use com_associations. Bussiness as usual.
#### Notes

One important note: There is still some joomla warnings that appears on the list view for items the table culdn't be loaded, but it as no impact can be solved later. There is already a PR in the core for solving that.

Now this works also with category "sections" (ex: com_mycomponent.myitemtype).
To get this working with section one must put a protect var in the model `categoryContext`. If var doesn't exist it fallbacks to componet (ex: com_mycomponent)

example:

``` php
/**
 * Categories context.
 *
 * @var    string
 * @since  __DEPLOY_VERSION__
*/
protected $categoryContext= 'com_mycomponent.myitemtype';
```

Also **PLEASE DON'T MERGE ANY OTHER PR UNTIL THIS ONE IS MERGED**.


andrepereiradasilva:
merged as decided

alikon:
:+1:

PR #92, opened 2016-08-13
Flexible db fields (redo - again)

andrepereiradasilva:
Flexible db fields.

Improves:
#### Test

All items in com_associations list views work as before.
#### 3º Party Support

If 3º party use the same fields as defined doesn't to do anything.
If 3º party does not use the same fields, needs to add `_columnAlias` property to his JTable override.


roland-d:
@andrepereiradasilva Your question:

> one question i have, is this the correct order to do the column alias?

This depends which column you want ;) The format should be:
old name -> new name

One thing I wonder is, why are there hard-coded field names in this change? I tried to understand the code but not really getting it. Could you explain this bit?

andrepereiradasilva:
the "hardcoded" fields

``` php
$fields = array(
    'id',
    'title',
    'alias',
    'ordering',
    'menutype',
    'level',
    'catid',
    'language',
    'access',
    'state',
    'created_user_id',
    'checked_out',
    'checked_out_time',
);
```

Are the db fields com_associations needs from the component item types to work properly.
(id, title, alias and language are mandatory)

The problem is many times each component item types uses it's own field names and not this ones.
The core itself does this (just compare com_content and com_categories db tables field names).

So in order to normalize this (so we can use in com_associations) we need to use the JTable _columnAlias property. And that's the motive for this change.

andrepereiradasilva:
ok just saw this https://docs.joomla.org/Column_alias

so, it's ok

andrepereiradasilva:
i'm gone a merge this an submit a PR to the core.

PR #97, opened 2016-08-14
Moving side by side css to the media folder

infograf768:
To test, check source 
simple way to test RTL: change en-GB.xml:
`<rtl>1</rtl>`

### DIFF
--- a/administrator/components/com_associations/views/association/tmpl/edit.php
+++ b/administrator/components/com_associations/views/association/tmpl/edit.php
@@ -15,78 +15,13 @@ JHtml::_('formbehavior.chosen', 'select');
 
 JHtml::_('script', 'com_associations/sidebyside.js', false, true);
 
-if (JFactory::getLanguage()->isRtl()) {
-	$this->app->getDocument()->addStyleDeclaration('
-
-		.sidebyside .outer-panel {
-			float: right;
-			width: 50%;
-		}
-		.sidebyside #right-panel .inner-panel {
-			border-right: 1px solid #999999 !important;
-		}
-		.sidebyside #left-panel .inner-panel {
-			padding-left: 10px;
-		}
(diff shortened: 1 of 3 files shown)


alikon:
work for me

andrepereiradasilva:
sorry guys but i don't understand why having a specifi rtl css ... when you do all ok that with just css.

you can just prepend the rtl classes overrides with 

``` css
html[dir=rtl]
```

example:

``` css
/* LTR (default) */
.sidebyside .outer-panel {
    float: left;
    width: 50%;
}

/* RTL (override) */
html[dir=rtl] .sidebyside .outer-panel {
    float: right;
}
```

infograf768:
@andrepereiradasilva 
Having a specific rtl css is what we always did and it makes things easier to read / and understand.
- I never heard about this kind of markup...
  `html[dir=rtl]`

andrepereiradasilva:
See https://github.com/joomla/joomla-cms/search?q=html%5Bdir%3Drtl%5D

infograf768:
Can we keep the old way please? ❕

andrepereiradasilva:
of course we can.

But is not IMHO the most correct way of doing thing.
for instance, this means if you changes the `.sidebyside #right-panel .inner-panel` class to have a border of 2px, you have to change in two files. 

This way we just change in one file ... so is more efficient.
For me is clearly is a better option. but if you insist in doing this way go ahead.

infograf768:
http://rishida.net/blog/?p=671#comment-10058

infograf768:
I really prefer the good old way. No chance to mistake.

andrepereiradasilva:
My opinion is that `old way` is clearly less eficient and harder to mantain and more mistake prone (ex: forget to change one the files when making a change ...).

But ... i will not discuss this anymore. If you want to merge it like this. Go ahead.

PR #98, opened 2016-08-14
Moving side by side css to css file

andrepereiradasilva:
#### Summary of Changes

Moves inline CSS to a CSS file.
#### Testing Instructions

CSS in LTR and RTL languages work fine.


infograf768:
Merged, but we may need some js to auto-resize the Iframe(s) (or get scroll bars) as this is unusable on mobiles when many fields.

andrepereiradasilva:
you're tottally right @infograf768 
i will make a new issue for that so joao can check

Issue #96, opened 2016-08-14
Associations tab not readonly for menu item reference (side by side)


infograf768:
As title says.

alikon:
i'm able to replicate

alikon:
if we apply https://github.com/joomla/joomla-cms/pull/11471 no more bug
i close this

andrepereiradasilva:
i'm going to reoepn this because it was expect to work without https://github.com/joomla/joomla-cms/pull/11471

andrepereiradasilva:
@jreys one more bug to solve

andrepereiradasilva:
closing as we have a PR to test https://github.com/joomla-projects/gsoc16_improved-multi-lingual/issues/102

PR #102, opened 2016-08-15
	Fixing associations tab not readonly on chzn fields

jreys:
Pull Request for Issue #96 .

### DIFF
--- a/media/com_associations/js/sidebyside.js
+++ b/media/com_associations/js/sidebyside.js
@@ -88,6 +88,7 @@ jQuery(document).ready(function($) {
 			// We need to re-enable the language field to save.
 
 			$('#' + task + '-association').contents().find('#jform_language').attr('disabled', false);
+			$('#' + task + '-association').contents().find('.chzn-done').attr('disabled', false);
 			window.frames[task + '-association'].Joomla.submitbutton(document.getElementById('adminForm').getAttribute('data-associatedview') + '.apply');
 		}
 
@@ -158,9 +159,6 @@ jQuery(document).ready(function($) {
 		// Remove modal buttons on the reference
 		reference.find('#associations').find('.btn').remove();
 		
-		// Make chzn boxes readonly on associations tab
-		reference.find('#associations').find('.chzn-container-single').addClass('chzn-disabled');
-
 		var parse = '';
 
 		$('#jform_itemlanguage option').each(function()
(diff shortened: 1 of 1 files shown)


infograf768:
This is not perfect here. The new association is not displayed in the reference as it is  for other items.
screen shot 2016-08-15 at 18 08 39

infograf768:
Now the association displays in the Target

and clear an assoc in the Target does not clear it from reference.

jreys:
The associations is displaying in the target in menus only, if I add [here](https://github.com/joomla-projects/gsoc16_improved-multi-lingual/pull/102/files#diff-329fb4d5c93171eac6542fd512f59d75R208) 

``` js
console.log($(this).contents().find('#details').hasClass('active'));
```

it returns true on menus, even when associations tab/content is active and false on other items, what I want here is to check if #associations is active, hence 

``` js
$(this).contents().find('#associations').hasClass('active')
```

false on menus, true on other items, I'm completely lost on this one, why is menus always a different case...?

infograf768:
Can you try this:
```php
diff --git a/administrator/components/com_associations/views/association/tmpl/edit.php b/administrator/components/com_associations/views/association/tmpl/edit.php
index 09aa561..5dae417 100644
--- a/administrator/components/com_associations/views/association/tmpl/edit.php
+++ b/administrator/components/com_associations/views/association/tmpl/edit.php
@@ -24,5 +24,5 @@
 		);
 ?>
-<button id="toogle-left-panel" class="btn btn-small" 
+<button id="toogle-left-panel" class="btn btn-small"
 		data-show-reference="<?php echo JText::_('COM_ASSOCIATIONS_EDIT_SHOW_REFERENCE'); ?>"
 		data-hide-reference="<?php echo JText::_('COM_ASSOCIATIONS_EDIT_HIDE_REFERENCE'); ?>"><?php echo JText::_('COM_ASSOCIATIONS_EDIT_HIDE_REFERENCE'); ?>
@@ -38,5 +38,5 @@
 				<iframe id="reference-association" name="reference-association"
 					src="<?php echo JRoute::_($this->editUri . '&task=' . $this->itemType->item . '.edit&id=' . (int) $this->referenceId); ?>"
-					height="100%" width="400px" scrolling="no"
+					height="100%" width="400px" overflow-y="auto"
 					data-action="edit"
 					data-item="<?php echo $this->itemType->item; ?>"
@@ -55,5 +55,5 @@
 				<iframe id="target-association" name="target-association"
 					src="<?php echo $this->defaultTargetSrc; ?>"
-					height="100%" width="400px" scrolling="no"
+					height="100%" width="400px" overflow-y="auto"
 					data-action="<?php echo $this->targetAction; ?>"
 					data-item="<?php echo $this->itemType->item; ?>"
```

If this is fine, we can get rid of the border.
It does not prevent from using media queries to push the target under the Reference on smaller screens.

What do you think?

alikon:
**where we are on this one ?**
stillbuggy

is still buggy

andrepereiradasilva:
@jreys please respond to @alikon

jreys:
I've wrote my problem [here](https://github.com/joomla-projects/gsoc16_improved-multi-lingual/pull/102#issuecomment-239866356), I can further explain it if you like.

Issue #94, opened 2016-08-13
tasks for 2016/08/14 - end


andrepereiradasilva:
Main Task for last week of reviewed timeline + task to do from other weeks:
#### From previous weeks
- [x] sidebyside: Implement copy feature by auto-filling fields with reference data (see also #8)
- [x] list view: [Proposal] Map relevant columns in component item model or table (more info #22)
- [ ] Implement "wiki" page with all 3º party requirements. (see https://github.com/joomla-projects/gsoc16_improved-multi-lingual/issues/21)
- [x] Add "Component: Item Type" title to views
- [ ] Create sidebyside-uncompressed.js
- [ ] Solve issues in github "issues"
#### From current week
- [ ] Implement "help" content.
- [x] Language strings need to be alpha ordered in ini files.
- [ ] Javascript/PHP Error catching
- [ ] Graphical and usability review
- [ ] Final tests and corrections

jreys:
- Add title to list view

andrepereiradasilva:
IMO this is not good we are only in 3 out 11 completed ... few days left ...

Issue #100, opened 2016-08-15
CSS Mobile and auto resize


andrepereiradasilva:
> we may need some js to auto-resize the Iframe(s) (or get scroll bars) as this is unusable on mobiles when many fields.

Side by side view must be reviewed to work on mobiles devices (tablets, smartphones, etc).
Needs to use a responsive approach with css media queries.

Hint: an easy way to test is to reduce browser width/height.

infograf768:
@jreys 
See this movie where window scrollbars are set to maximum
https://www.dropbox.com/s/2ci79ci2ou22lbx/mobileissues.mp4?dl=0

infograf768:
Can you try this:

``` php
diff --git a/administrator/components/com_associations/views/association/tmpl/edit.php b/administrator/components/com_associations/views/association/tmpl/edit.php
index 09aa561..5dae417 100644
--- a/administrator/components/com_associations/views/association/tmpl/edit.php
+++ b/administrator/components/com_associations/views/association/tmpl/edit.php
@@ -24,5 +24,5 @@
        );
 ?>
-<button id="toogle-left-panel" class="btn btn-small" 
+<button id="toogle-left-panel" class="btn btn-small"
        data-show-reference="<?php echo JText::_('COM_ASSOCIATIONS_EDIT_SHOW_REFERENCE'); ?>"
        data-hide-reference="<?php echo JText::_('COM_ASSOCIATIONS_EDIT_HIDE_REFERENCE'); ?>"><?php echo JText::_('COM_ASSOCIATIONS_EDIT_HIDE_REFERENCE'); ?>
@@ -38,5 +38,5 @@
                <iframe id="reference-association" name="reference-association"
                    src="<?php echo JRoute::_($this->editUri . '&task=' . $this->itemType->item . '.edit&id=' . (int) $this->referenceId); ?>"
-                   height="100%" width="400px" scrolling="no"
+                   height="100%" width="400px" overflow-y="auto"
                    data-action="edit"
                    data-item="<?php echo $this->itemType->item; ?>"
@@ -55,5 +55,5 @@
                <iframe id="target-association" name="target-association"
                    src="<?php echo $this->defaultTargetSrc; ?>"
-                   height="100%" width="400px" scrolling="no"
+                   height="100%" width="400px" overflow-y="auto"
                    data-action="<?php echo $this->targetAction; ?>"
                    data-item="<?php echo $this->itemType->item; ?>"
```

[...]

PR #104, opened 2016-08-15
Adding copy button action

jreys:
Pull Request for Issue #8 and #94.
#### Summary of Changes

Copy all now working, test to see if this is a good way, IMO it's slow but it's the only way to do save & copy and getting the new ID.

Copy field by field not yet implemented
#### Testing Instructions

Check if copy all is working properly, it will only show up when there's a language selected

### DIFF
--- a/administrator/components/com_associations/views/association/tmpl/edit.php
+++ b/administrator/components/com_associations/views/association/tmpl/edit.php
@@ -41,7 +41,8 @@ $options = array(
 					data-action="edit"
 					data-item="<?php echo $this->itemType->item; ?>"
 					data-id="<?php echo $this->referenceId; ?>"
-					data-language="<?php echo $this->referenceLanguage; ?>">
+					data-language="<?php echo $this->referenceLanguage; ?>"
+					data-editurl="<?php echo JRoute::_($this->editUri); ?>">
 				</iframe>
 			</div>
 		</div>
--- a/media/com_associations/js/sidebyside.js
+++ b/media/com_associations/js/sidebyside.js
@@ -240,22 +240,21 @@ jQuery(document).ready(function($) {
 					// Update main frame data-id attribute (used after save).
 					this.setAttribute('data-id', targetLoadedId);
 					this.setAttribute('data-action', 'edit');
-
-					// Update the reference item associations tab.
-					var reference     = document.getElementById('reference-association');
-					var languageCode  = targetLanguage.replace(/-/, '_');
-					var title         = $(this).contents().find('#jform_title').val()
-
-					// - For modal association selectors.
-					$(reference).contents().find('#jform_associations_' + languageCode + '_id').val(targetLoadedId);
-					$(reference).contents().find('#jform_associations_' + languageCode + '_name').val(title);
-
-					// - For chosen association selectors (menus).
-					$(reference).contents().find('#jform_associations_' + languageCode + '_chzn').remove();
-					$(reference).contents().find('#jform_associations_' + languageCode).append('<option value=\"'+ targetLoadedId + '\">' + title + '</option>');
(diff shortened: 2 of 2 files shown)


infograf768:
Tested with articles. Looks like working here.
will test others later.

infograf768:
Tested categories. I see only one small issue: when changing title of target after copy and then save target, the association is not updated in the reference (but it is in manager). It is only updated in reference when saving reference.

Screeenshots: Reference is 

> subsubcat it-IT



Going side by side
 

Copy reference


Reference displays new association


**Changing Target title and saving: reference is not updated**

5_aftersavemanager

Manager shows new name fine and items are already associated
6-manager

reference is updated after save
aftersavereference

NOTE: after saving reference the scroll bars disappear. I got them first by modifying edit.php
See: https://github.com/joomla-projects/gsoc16_improved-multi-lingual/issues/100#issuecomment-240317284

infograf768:
Copying reference menu items gives:
[17-Aug-2016 07:15:41 UTC] PHP Notice:  Undefined index: associations in ROOT/administrator/components/com_menus/models/item.php on line 1384

Copying reference category gives
[17-Aug-2016 07:17:33 UTC] PHP Notice:  Undefined index: associations in ROOT/administrator/components/com_categories/models/category.php on line 575

infograf768:
NOTE: A 3rd party component needs in its model a CORRECT  `generateNewTitle()` method for `Save as Copy` in order to copy reference, otherwise we get the error (here for Sermonspeaker:
`COM_SERMONSPEAKER_ERROR_ALIAS="This alias is already used for another item (remember it may be a trashed item)."`

infograf768:
We may have to enable in com_associations a custom generatenewtitle in case the 3rd party component does not implement Save as Copy (save2copy) correctly.

infograf768:
https://github.com/joomla-projects/gsoc16_improved-multi-lingual/pull/104#issuecomment-240333325
has PRs in core. One merged, the other waiting https://github.com/joomla/joomla-cms/pull/11664

EDIT: both now merged

infograf768:
@jreys 

Please try to solve **now** the issue I explained above: changing target title (and saving Target) does not change on the fly the association in the Reference item in the same view.

infograf768:
@jreys
Concerning the copying of items when there is no generatenewtitle for Save as Copy

I found out that if we change the alias of the reference, then the copy is done OK after some refreshing of the reference item. Otherwise we get the error BUT with the full admin menus in the Iframe

Can you propose a solution?

infograf768:
@jreys
When I use sermon speaker and save Reference or Target (not category), the admin menus show.
Any idea how to solve this or add something to the requirements?

screen shot 2016-08-25 at 08 44 52

The reason is they do not have the Edit part in their modal. Can we trace this to prevent these items from being taken care of in com_associations if we can't solve ourseleves?

Maybe check if Apply contains `&amp;tmpl=component` and add it if not?

andrepereiradasilva:
we already talked befora about this JM.
it needs to be a requirement.

infograf768:
The problem, Andre, is that we do not check (or can't check) and the way com_associations works, it will propose the concerned items to com_associations, even if not fully compatible.

infograf768:
Anyhow this surely can be solved
https://github.com/joomla-projects/gsoc16_improved-multi-lingual/pull/104#issuecomment-242291882

andrepereiradasilva:
JM, that i know of, we have no way of checking this in the model or table.
So it must be a requirement

infograf768:
tks, @jreys 
This has solved the issue of modifying the target name.

@andrepereiradasilva 
We can merge this if compatible with PLT requirements, i.e. if it has no effect on a change to the way we fetch the necessary informations for Items and table in a new helper.

andrepereiradasilva:
merged

Issue #107, opened 2016-08-16
BUG: Forcedlanguage lost after using Clear


infograf768:
See movie
https://www.dropbox.com/s/uzs4hqlwj4cvjro/forcedlanguagelost.mp4?dl=0

jreys:
This can now be closed as we have #108 to test.

PR #108, opened 2016-08-16
Forcedlanguage lost after using Clear FIX

jreys:
Pull Request for Issue #107
### Summary of Changes

When I was clearing association, I was giving the switcher reference language instead of target (for that position), i.e. if fr-FR had fr-FR:92:edit, when clearing it would replace with en-GB:0:add, everything was right except language.
### Testing Instructions

Clear an association and check if forced language on the target is still set to reference's forced language


infograf768:
Tested OK. Merged

Issue #101, opened 2016-08-15
[RFC] - fix ACL on update


alikon:
this snippet of code for clone from the com_content  "core.admin", "core.manage"  rules  when updating

``` php

$assetRules = JAccess::getAssetRules('com_content');
$assetData  = $assetRules->getData();
$rules = new JRegistry();           
$rules->loadString($assetData['core.admin']);  
$core_admin =$rules->__toString();
$rules = new JRegistry();           
$rules->loadString($assetData['core.manage']);  
$core_manage=$rules->__toString();

$cloned_rules = '{"core.admin":'.$core_admin.','. '"core.manage":'.$core_manage.'}';
//echo '{"core.admin":'.$core_admin.','. '"core.manage":'.$core_manage.'}';
```

so in  administrator/components/com_admin/script.php

``` php
if ($component === 'com_associations')
{
     $asset->rules = $cloned_rules';
}
```

for sure can be done more more better   ....
any feedback....

andrepereiradasilva:
so you are adding the same groups for Administrator and Manager as the exisitng for com_content? 

it could be one way to solve the issue.
Anyway i think you are doing that in all joomla updates right? So it will change on each update, or am i wrong?

@infograf768 @jreys your opinion here please.

alikon:
> so you are adding the same groups for Administrator and Manager as the exisitng for com_content?

yes

> Anyway i think you are doing that in all joomla updates right? So it will change on each update, or am i wrong?

wrong
it will be done only on the first update i.e only when the com_associations is installed for the 1st time

see
https://github.com/joomla-projects/gsoc16_improved-multi-lingual/blob/staging/administrator/components/com_admin/script.php#L1604

andrepereiradasilva:
ok then

andrepereiradasilva:
closed as we have a PR

Issue #93, opened 2016-08-13
Fix ACL problem on update


jreys:
#### Actual result

Right now when updating Joomla to support this new component, ACL is currently broken, it was fixed on install but on update the ACL might be different and we didn't find a general way to do this.
#### Expected result

Have update ACL == install ACL

andrepereiradasilva:
i didn't foun it interely is one idea that can be explored
is using the udpate script to dynamicly update the com_associations ACL
See https://github.com/infograf768/joomla-cms/pull/30/files#diff-e3c745afdbb43914c4f183a8c1587346R1619

alikon:
we need to add com_associations https://github.com/joomla-projects/gsoc16_improved-multi-lingual/blob/staging/administrator/components/com_admin/script.php#L1588
as well

alikon:
already done by my mistake direct commit instead of a new PR
forgive me :worried:

andrepereiradasilva:
np, for small things i do it too

andrepereiradasilva:
closing as we have a RFC to discuss it https://github.com/joomla-projects/gsoc16_improved-multi-lingual/issues/101

PR #109, opened 2016-08-16
fix ACL on update

alikon:
Pull Request for Issue #93.
### Summary of Changes

cloned rules from com_content when update
### Testing Instructions

check that ACL  when update have the same rules as for com_content for core.admin, core.manage

see #101

### DIFF
--- a/administrator/components/com_admin/script.php
+++ b/administrator/components/com_admin/script.php
@@ -1614,16 +1614,18 @@ class JoomlaInstallerScript
 
 			if ($component === 'com_associations')
 			{
-     				$assetRules = JAccess::getAssetRules('com_content');
+				// Clone from com_content
+				$assetRules = JAccess::getAssetRules('com_content');
 				$assetData  = $assetRules->getData();
-				$rules = new JRegistry();
+				$rules = new JRegistry;
 				$rules->loadString($assetData['core.admin']);
 				$core_admin = $rules->__toString();
-				$rules = new JRegistry();
+				$rules = new JRegistry;
 				$rules->loadString($assetData['core.manage']);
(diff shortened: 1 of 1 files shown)


andrepereiradasilva:
cs

andrepereiradasilva:
@alikon i'm still not enterrily convinced we should copy the com_content permissions, can't we create a access security issue with this?

isn't it better (or at least more safe) to check the user groups "Manager" and "Administrator" have the same ids as the install and if so use them, if not don't do anything?

alikon:
the most easy thing that we can do is do like other new core components behave on update (nothing) ....

andrepereiradasilva:
do you think we should ask the PLT an opinion for this? maybe there is another way?

alikon:
Why not we are open to listen more experienced people ;)

infograf768:
maybe contact the weblinks people as i discovered they indeed do nothing atm and my manager could not access to weblinks. Had to manually set it by a superuser.

andrepereiradasilva:
lol

alikon:
the com_weblinks as it is decoupled not more a core components, is not a correct example i think

andrepereiradasilva:
see michael comment https://github.com/joomla/joomla-cms/pull/11834/files/9ac28b8c2c14b7b6b79e963a5ea20fc0041b7dc2#r76787099

alikon:
if i've understand well micheal comment we already do this , we still need to decide if we want to clone rules from com_content or not

andrepereiradasilva:
IMHO we close this and do what all the other components do

alikon:
yes fine for me

andrepereiradasilva:
closing

PR #111, opened 2016-08-17
Making CSS responsive with smaller devices and adding scrollbars

jreys:
Pull Request for Issue #100 
### Testing Instructions

Resize window and check if the behaviour is intended

### DIFF
--- a/media/com_associations/css/sidebyside.css
+++ b/media/com_associations/css/sidebyside.css
@@ -9,9 +9,6 @@
 }
 
 .sidebyside #left-panel .inner-panel {
-	border-style: solid;
-	border-color: #999999;
-	border-width: 0 1px 0 0;
 	padding: 0 10px 0 0;
 }
 
@@ -31,6 +28,7 @@
 #reference-association, #target-association {
 	width: 100%;
 	height: 1500px;
+	overflow-y: auto;
 	border: 0 !important;
 }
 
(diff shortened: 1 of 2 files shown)


media/com_associations/css/sidebyside.css
```diff
@@ -9,9 +9,6 @@
 }
 
 .sidebyside #left-panel .inner-panel {
-	border-style: solid;
-	border-color: #999999;
-	border-width: 0 1px 0 0;
 	padding: 0 10px 0 0;
```
andrepereiradasilva: if you remove the border css styles here you should remove the border in `html[dir=rtl] .sidebyside #left-panel .inner-panel` too...

infograf768:
Works here.

andrepereiradasilva:
doesn't work here. i get an horizontal scroll bar

image

infograf768:
@andrepereiradasilva 
On which browser? Here on FF Mac, only the vertical bar

infograf768:
@andrepereiradasilva 
For me the horizontal scroolbar happens ONLY when some stuff is larger than the width.
It is the case when I display for example a menu item of type single article where we have the Select Article field very large (the EDit button goes over the edge)
As soon as I display for the same menu item the Options tab, the horizontal scroll bar disapears.
 screen shot 2016-08-17 at 20 40 01

So I think this PR is OK

PR #112, opened 2016-08-18
Adding context title to the list view

jreys:
Pull Request for Issue #94 -> Add "Component: Item Type" title to views
### Testing Instructions

Go to the list view and change item type, it should change in title, was also tested in a RTL language.


infograf768:
Works fine here.

The only aspect I don't like so much is Menus / Menus, but it is also the case in the Select Item Type selector.
I wonder if we should not make an exception for Menus and change to `Menu Items` and therefore get in the selector as well as in the page title `Menu / Menu Items` with the string
`COM_MENUS_SUBMENU_ITEMS="Menu Items"`

jreys:
I think with this new title, we can replace itemType's title with Items again, so it will be:
Menus / Items, no need to hardcode for menus, just change that in the helper

infograf768:
We would not like change what we get for example for sermonspeaker (where we get the right title)

screen shot 2016-08-18 at 12 30 08

jreys:
the change I proposed is only regarding menus, the rest is fine

andrepereiradasilva:
the string used is `COM_MENUS_ITEMS`

is from here: https://github.com/joomla-projects/gsoc16_improved-multi-lingual/blob/staging/administrator/components/com_associations/helpers/associations.php#L208

actually this code could be better (use plural forms) instead of `. 'S'`

andrepereiradasilva:
@infograf768 can you check that?

infograf768:
> actually this code could be better (use plural forms) instead of . 'S'

Not sure I get it. We are not in a case where we can use JText::plural

But we can make it a requirement (and propose another PR to core...) with a special string constant for each component concerned.
In our case (Menus) we do not have that string:
COM_MENUS_ITEMS="Menu items"

infograf768:
I suggest to add the string above (COM_MENUS_ITEMS="Menu items") for menus in this PR.

andrepereiradasilva:
For instance if you have an Item type called SILLY it will look for a string called SILLYS and not SILLIES as it should.

how do we solve that?

andrepereiradasilva:
@infograf768 please chek this com_finder helper funcion: https://github.com/joomla/joomla-cms/blob/staging/administrator/components/com_finder/helpers/language.php#L28

administrator/language/en-GB/en-GB.com_associations.ini
```diff
@@ -25,8 +25,9 @@ COM_ASSOCIATIONS_FILTER_MENUTYPE_LABEL="Menu"
 COM_ASSOCIATIONS_FILTER_SEARCH_DESC="Search an item by it's title"
 COM_ASSOCIATIONS_FILTER_SEARCH_LABEL="Search item"
 COM_ASSOCIATIONS_FILTER_SELECT_ITEM_TYPE="- Select Item Type -"
-COM_ASSOCIATIONS_HEADER_EDIT="Multilingual Associations: Associations"
-COM_ASSOCIATIONS_HEADER_SELECT_REFERENCE="Multilingual Associations: Select Reference"
+COM_ASSOCIATIONS_HEADER_EDIT="Multilingual Associations: Edit Associations (%1s =&gt; %2s)"
+COM_ASSOCIATIONS_HEADER_SELECT_FILTERS="Multilingual Associations: Select Item Type and Language"
+COM_ASSOCIATIONS_HEADER_SELECT_REFERENCE="Multilingual Associations: Select Reference (%1s =&gt; %2s)"
```
andrepereiradasilva: remove the ": Select Reference" here please

administrator/language/en-GB/en-GB.com_menus.ini
```diff
@@ -123,6 +123,7 @@ COM_MENUS_ITEM_IS_DEFAULT="Is default"
 COM_MENUS_ITEM_MODULE_ASSIGNMENT="Module Assignment"
 COM_MENUS_ITEM_REQUIRED="Required"
 COM_MENUS_ITEM_ROOT="Menu Item Root"
+COM_MENUS_ITEMS="Menu items"
```
andrepereiradasilva: this is a core patch shouldn't be here

administrator/components/com_associations/views/associations/view.html.php
```diff
@@ -163,9 +163,17 @@ public function display($tpl = null)
 	 */
 	protected function addToolbar()
 	{
-		$user  = JFactory::getUser();
+		$user    = JFactory::getUser();
```
andrepereiradasilva: why the spaces?

jreys: didn't remove them, was aligning with old variable, will remove them

administrator/language/en-GB/en-GB.com_associations.ini
```diff
@@ -25,8 +25,9 @@ COM_ASSOCIATIONS_FILTER_MENUTYPE_LABEL="Menu"
...
 COM_ASSOCIATIONS_FILTER_SEARCH_LABEL="Search item"
 COM_ASSOCIATIONS_FILTER_SELECT_ITEM_TYPE="- Select Item Type -"
-COM_ASSOCIATIONS_HEADER_EDIT="Multilingual Associations: Associations"
-COM_ASSOCIATIONS_HEADER_SELECT_REFERENCE="Multilingual Associations: Select Reference"
+COM_ASSOCIATIONS_HEADER_EDIT="Multilingual Associations: Edit Associations (%1s =&gt; %2s)"
+COM_ASSOCIATIONS_HEADER_SELECT_FILTERS="Multilingual Associations: Select Item Type and Language"
+COM_ASSOCIATIONS_HEADER_SELECT_REFERENCE="Multilingual Associations (%1s =&gt; %2s)"
 COM_ASSOCIATIONS_HEADING_ASSOCIATION="Associations"
 COM_ASSOCIATIONS_HEADING_MENUTYPE="Menu"
 COM_ASSOCIATIONS_HEADING_MENUTYPE_ASC="Menu ascending"
```
andrepereiradasilva, line 30: why are you changing the file permissions to 755 ?

administrator/language/en-GB/en-GB.com_associations.ini
```diff
@@ -25,8 +25,9 @@ COM_ASSOCIATIONS_FILTER_MENUTYPE_LABEL="Menu"
 COM_ASSOCIATIONS_FILTER_SEARCH_DESC="Search an item by it's title"
 COM_ASSOCIATIONS_FILTER_SEARCH_LABEL="Search item"
 COM_ASSOCIATIONS_FILTER_SELECT_ITEM_TYPE="- Select Item Type -"
-COM_ASSOCIATIONS_HEADER_EDIT="Multilingual Associations: Associations"
-COM_ASSOCIATIONS_HEADER_SELECT_REFERENCE="Multilingual Associations: Select Reference"
+COM_ASSOCIATIONS_HEADER_EDIT="Multilingual Associations: Edit Associations (%1s =&gt; %2s)"
+COM_ASSOCIATIONS_HEADER_SELECT_FILTERS="Multilingual Associations: Select Item Type and Language"
+COM_ASSOCIATIONS_HEADER_SELECT_REFERENCE="Multilingual Associations (%1s =&gt; %2s)"
```
andrepereiradasilva: maybe better?:

```
COM_ASSOCIATIONS_TITLE_EDIT="Multilingual Associations: Edit Associations (%1s &gt; %2s)"
COM_ASSOCIATIONS_TITLE_LIST="Multilingual Associations (%1s &gt; %2s)"
COM_ASSOCIATIONS_TITLE_LIST_SELECT="Multilingual Associations: Select Item Type and Language"
```

administrator/language/en-GB/en-GB.com_associations.ini
jreys, line 30: Didn't know git stored permissions

administrator/language/en-GB/en-GB.com_associations.ini
```diff
@@ -45,4 +43,7 @@ COM_ASSOCIATIONS_SELECT_MENU="- Select Menu -"
 COM_ASSOCIATIONS_SELECT_TARGET="Select Target"
 COM_ASSOCIATIONS_SELECT_TARGET_LANGUAGE="- Select Target Language -"
 COM_ASSOCIATIONS_TITLE="Associations"
+COM_ASSOCIATIONS_TITLE_EDIT="Multilingual Associations: Edit Associations (%1s &gt; %2s)"
+COM_ASSOCIATIONS_TITLE_LIST="Multilingual Associations (%1s &gt; %2s)"
+COM_ASSOCIATIONS_TITLE_LIST_SELECT="Multilingual Associations: Select Item Type and Language"
```
andrepereiradasilva: if you change here ... don't you need to change elsewhere ... ?

andrepereiradasilva:
merged

Issue #114, opened 2016-08-26
Discuss component arquitecture


andrepereiradasilva:
Placeholder to discuss component arquitecture in the light of the comments in https://github.com/joomla/joomla-cms/pull/11623.

infograf768:
@jreys 
I think we should go with the new helper in the extension back-end.
First aspect to define: what exactly should it contain that would be fetched by com_associations own helper to include the extension in com_associations?
I mean the minimum required and some stuff that would or not exist which would change some UI in com_associations (like save2copy with a generatenewtitle() for example)

alikon:
@jreys for reference look at how com_categories do for manage items count https://github.com/joomla/joomla-cms/pull/9429
or https://github.com/joomla/joomla-cms/pull/10895 do for tags

andrepereiradasilva:
For reference, any change made:
- needs to support component multi-item types per component
- needs to support component multi-item types categories per component
- needs to respect component multi-item types ACL
- needs to respect component multi-item types checkin/checkout

PR #116, opened 2016-10-02
Adding initial helpers for contacts, content, menus and newsfeeds

jreys:
Pull Request for Issue #114
### Summary of Changes

Added helpers on each component like stated [here](https://github.com/joomla/joomla-cms/pull/11623#issuecomment-242345372) added an extra function for categories, it helps imo.

I would like you to review to check if it's the correct way of doing this or if I'm doing something wrong, thanks!
### Testing Instructions

None yet.


infograf768:
Looks like the way to go. 
Would I guess need further code to define $typeAlias accepted (there may be multiple views) as well as Categories.
Example for SermonSpeaker:

```
public static function getAssociations($id = 0, $view = null)
    {
        jimport('helper.route', JPATH_COMPONENT_SITE);

        $jinput = JFactory::getApplication()->input;
        $view   = is_null($view) ? $jinput->get('view') : $view;
        $id     = empty($id) ? $jinput->getInt('id') : $id;

        switch ($view)
        {
            case 'sermon':
                return self::_findItems($id, $view, '#__sermon_sermons');
            case 'serie':
                return self::_findItems($id, $view, '#__sermon_series');
            case 'speaker':
                return self::_findItems($id, $view, '#__sermon_speakers');
            case 'category':
            case 'categories':
                return self::getCategoryAssociations($id, 'com_sermonspeaker');
        }

        return array();
    }
```

Compare to com_content:

```
    public static function getAssociations($id = 0, $view = null)
    {
        $jinput = JFactory::getApplication()->input;
        $view   = is_null($view) ? $jinput->get('view') : $view;
        $id     = empty($id) ? $jinput->getInt('id') : $id;

        if ($view == 'article' || $view == 'category' || $view == 'featured')
        {
            if ($id)
            {
                $associations = JLanguageAssociations::getAssociations('com_content', '#__content', 'com_content.item', $id);

                $return = array();

                foreach ($associations as $tag => $item)
                {
                    $return[$tag] = ContentHelperRoute::getArticleRoute($item->id, (int) $item->catid, $item->language);
                }

                return $return;
            }
        }

        if ($view == 'category' || $view == 'categories')
        {
            return self::getCategoryAssociations($id, 'com_content');
        }

        return array();
    }
```

Can you now add the necessary changes to fetch these values in com_associations?

jreys:
Yes the plan was to fetch on the com_associations' helper. Also, that change you proposed, should be on a helper method listing all views with associations? (in an array or something like that)

infograf768:
Basically, I was thinking we need to get in this new helper some variables we got in com_associations/helpers/associations.php from the model(s) of the component which is implementing associations.
@Bakual did not like we get `associationsContext` and `categoryContext` from the model for example.
Same for `typeAlias`.

@andrepereiradasilva 
What do you think?

infograf768:
@jreys 
Nothing new?

jreys:
Updated! Somehow I'm doing something wrong, tests show correctly but on a 3dp I don't think it will work at 100%, any insights?

infograf768:
Testing now.

infograf768:
I can't even test here, @jreys 
Fatal error: Class 'ContactAssociationsHelper' not found in /Applications/MAMP/htdocs/gsocmulti/administrator/components/com_associations/helpers/associations.php on line 100

jreys:
My bad, updated locally and not on the repo, should now work as intended

infograf768:
An associations helper is missing for weblinks which also creates a fatal error. As weblinks may or may not be installed, we should check if the new helper is present and if not, do not throw a fatal error and rather do not display the component in the list of types choice.

Suggestion, line 98 of /Applications/MAMP/htdocs/gsocmulti/administrator/components/com_associations/helpers/associations.php.

``` php
            if (file_exists($helpersPath . '/associations.php') && $it[$key]->item !== 'category')
            {
```

This may also have to be checked in /administrator/components/com_associations/models/fields/itemtype.php —although, in the case of weblinks we only deal with categories— if we enforce absolutely the new helper.

Also, `public static function getColumnTableAlias()` is not used which creates undefined index issues in com_contacts as 'state' is in fact 'published' in that table.

I also found some other stuff to modify to avoid registry errors. One has to add `isset(`
In view/default.php
Change to:

``` php
                    <?php if (isset($this->itemType->fields->state) && !is_null($this->itemType->fields->state)) : ?>
                        <td class="center">
                            <span class="<?php echo $iconStates[$this->escape($item->state)]; ?>"></span>
                        </td>
                    <?php endif; ?>
```

Similar error in itemtype.php field

``` php
                    if ($itemType->associations->support && !in_array($itemType->assetKey, $itemTypeList))
                    {
                        if (isset($itemType->title) && $itemType->title)
                        {
                            $options[$itemType->componentTitle][] = JHtml::_('select.option', $itemType->assetKey, $itemType->title);
                        }

                        array_push($itemTypeList, $itemType->assetKey);
                    }
```

Note: I added a simple associations helper in sermonspeaker and it looks like working without specifying types (we have 3 types and one category for that component.)
This component uses the default array for columns names for its 3 tables which may explain that.
A solution has to be found to cope with possible differences between type specific tables not using same column names.

infograf768:
@jreys 

???

jreys:
Ok added those checks, I have some questions now regarding column alias, should we force 3dp to have a table to get column alias or should I add on the individual helpers all the fields that should be fetched? i.e. having for example on contacts' helper all of the fields that contact should fetch:
- id, name, alias, ordering, menutype, level, catid, language, access, state, created_by, checked_out, checked_out_time

infograf768:
The new helpers imho should provide all necessary.
Then you should use the new function `public static function getColumnTableAlias()` to get the right values in com_associations helper.
Also, as I said above:
`@Bakual did not like we get associationsContext and categoryContext from the model for example.
Same for typeAlias.`
Therefore I think these values have to be fetched from the new helpers.

jreys:
Ok then, I agree, it's better that way

andrepereiradasilva:
Imho the helpers should be one per item type. Example helpers/associations/article.php. why? Because you can have association per item type in a component. Also each one of those can have categories.

jreys:
So if you have `../helpers/associations/categories.php` the item supports category associations?

infograf768:
I think @andrepereiradasilva means that for each type helper, you would anyway get

``` php
    /**
     * Check if item supports category associations
     *
     * @return  boolean
     *
     * @since  __DEPLOY_VERSION__
     */
    public static function hasAssociationsCategories()
    {
        return true;
    }
```

AND a new method with the categories concerned.

andrepereiradasilva:
Yes, that's What i mean.

For instance, imagine banners have associations in banners and clients item types, you would have
image

``` php
helpers/associations/banner.php
helpers/associations/client.php
```

Since tracks does not ahave associations it does not need the helper.
Reggarding the categories, only helpers/associations/banner.php will have support for it since only banner item type supports categories

administrator/components/com_associations/helpers/associations.php
```diff
@@ -89,8 +90,22 @@ public static function getItemTypeProperties($key = '')
...
 			$it[$key]->model = JModelLegacy::getInstance($itemName, $componentName . 'Model', array('ignore_request' => true));
 
-			// If component item type model cannot loaded, or associations properties does not exist, item type does not support associations.
-			$it[$key]->associations->support = $it[$key]->model && $it[$key]->model->get('associationsContext');
+			// Load component's helper
+			$helperName = $componentName . 'AssociationsHelper';
+			JLoader::register($helperName, $helpersPath . '/associations.php');
+
+			// If component item helper cannot loaded, or hasAssociationsSupport is false, item type does not support associations.
+			if (file_exists($helpersPath . '/associations.php') && $it[$key]->item !== 'category')
```
andrepereiradasilva: you register it with JLoader before and only after you check if the file exists? doen't make a lot of sense to me, please check.

administrator/components/com_associations/helpers/associations.php
```diff
@@ -89,8 +90,22 @@ public static function getItemTypeProperties($key = '')
...
+			{
+				$helper = new $helperName;
+
+				$it[$key]->associations->support           = $helper->hasAssociationsSupport();
+				$it[$key]->associations->supportCategories = $helper->hasAssociationsCategories();
```
andrepereiradasilva: you should check before if class is loaded and if those method exists ...

administrator/components/com_associations/helpers/associations.php
```diff
@@ -89,8 +90,22 @@ public static function getItemTypeProperties($key = '')
...
+			}
+			else
+			{
+				$it[$key]->associations->support = true;
```
andrepereiradasilva: so if no helper exists ... it supports associations?

jreys:
Ok, now the fields are ok with the respective aliases and all, now I have some questions and remarks:
- How to deal with categories? creating a `com_categories/helpers/associations/category.php`?
- I'll remove all model/table related stuff after everything is fine so I can keep testing with the old way
- I've made a check to not include categories since I don't know yet how to handle them, nvm that please

Thanks

administrator/components/com_contact/helpers/associations/contact.php
```diff
@@ -0,0 +1,68 @@
...
+			'id'               => 'id',
+			'title'            => 'name',
+			'alias'            => 'alias',
+			'ordering'         => 'ordering',
+			'menutype'         => null,
+			'level'            => null,
+			'catid'            => 'catid',
+			'language'         => 'language',
+			'access'           => 'access',
+			'state'            => 'state',
```
infograf768: should be 'published'

infograf768:
`PHP Notice:  Undefined property: Joomla\Registry\Registry::$fields` everywhere in administrator/components/com_associations/helpers/associations.php

I guess you should use `!isset` instead of `is_null`

infograf768:
Something like

``` php
            if (!isset($it[$key]->fields->catid))
            {
                array_push($it[$key]->excludeOrdering, 'category_title');
            }
            if (!isset($it[$key]->fields->menutype))
            {
                array_push($it[$key]->excludeOrdering, 'menutype_title');
            }
            if (!isset($it[$key]->fields->access))
            {
                array_push($it[$key]->excludeOrdering, 'access_level');
            }
            if (!isset($it[$key]->fields->ordering))
            {
                array_push($it[$key]->excludeOrdering, 'ordering');
            }

            // Check the default ordering (ordering is the default. If not supported, fallback to title).
            $it[$key]->defaultOrdering = !isset($it[$key]->fields->ordering) ? array('title', 'ASC') : array('ordering', 'ASC');

            // If item does not have id, title, alias and language, it cannot support associations.
            if (!isset($it[$key]->fields->id) || !isset($it[$key]->fields->title) || !isset($it[$key]->fields->alias) || !isset($it[$key]->fields->language))
            {
                $it[$key]->associations->support = false;
            }
```

infograf768:
Concerning categories:
As I see it we could use a general com_categories/helpers/associations/category.php to get the table column alias.

Then in the new helpers per type, add a specific function returning the category extension.

In our core cases the array would contain only one category.
For sermonspeaker, although we have 3 types, it is the same category which is used.

if sermonspeaker had one category per type, we would have:
com_sermonspeaker.serie, com_sermonspeaker.speaker, com_sermonspeaker.sermon

Bakual:
I just looked at the current code and tried to reverse engineer what you do to get the item types for the dropdown in the associations listing view. You still traverse each component and each model, check if component is enabled and instantiate each model only to get the list of available item types. The types itself are guessed based on component.model name. And that is only the dropdown...
When it comes to generating the list itself, the associations models needs information from the 3rd party model and table class (fields, mappings, database table, ...) to run its own query. That model alone is very complex as it tries to do everyhing automagically. It has hardocded stuff in it relevant only for specific core extensions (eg menu stuff). It will fail on every 3rd party extension that doesn't follow "core" standards (eg access viewlevels only done on category level instead on item level).

The idea with the helper certainly is the better direction to go, but it has to go much further. Eg the list of available item types need to be provided by that helper class. You would still have to traverse the components directories (which imho still isn't nice) but only look for the helper class and run the proper method in it. This method has to return the available item types for the dropdown.
The list itself has to be fetched by the 3rd party model, preferable not instantiating it from com_associations (you may miss some required prerequisites and generate fatal errors) but through the helper. 

You could also use a plugin approach. Then you just run a plugin event (eg "getAssociationItemTypes") to fetch the needed information. Imho that would be a better approach as it would allow the user customise the behaviour of com_associations. Similar to how com_search and com_finder works.
Each component then would need one or multiple plugins (eg one per type) in a newly created plugin group (eg associations).

Just find a way so you don't have to directly fetch any data and instead request it from the 3rd party somehow, either though a helper (no traversing/guessing of filenames here - use a standardised name!) or plugin events. Everything else is wrong.

jreys:
@Bakual , thanks for your insights on this, so you think the best way is to add a file always in the same place (eg. com_xpto/helpers/associations.php) and that file has the sole responsibility of giving every info needed, if it supports associations, categories, which tables are affected, etc.) ? That looks better and in case the extension doesn't have that file it does not support associations and won't be listed.

alikon:
@jreys see #9429 and/or #10895 they use quite the same approach

infograf768:
If I follow @Bakual idea, I guess he would prefer plugins to helper (similar to the finder plugins).
The reason behind this is that com_associations would not have to search for a helper in all components installed on the site, but just the existing specific plugins in a new plugin folder `associations`from which it would fetch all infos and links to the necessary methods in the component concerned by the plugin.

Bakual:
> that file has the sole responsibility of giving every info needed, if it supports associations, categories, which tables are affected,

Exactly, minus the "which tables are affected". Let it create the query for you and you don't need any table knowledge at all (and no field mapping or anything).

>  I guess he would prefer plugins to helper

Plugins are great to get the list of item types as com_associations could just shout to the system and get the response from the plugins. No need to traverse the directories and search for the existence of files. Helpers are ideal when you only deal with one extension and already know which extension you're dealing with (eg fetch an item from that extension). Since we will have to do both type of data fetching, there are points for both ways. I wouldn't do both a plugin **and** a helper, so both ways would be fine imho :)

jreys:
So  I should keep with the helper approach because learning how to create a plugin takes longer than expected (I think).

> Exactly, minus the "which tables are affected". Let it create the query for you and you don't need any table knowledge at all (and no field mapping or anything).

How do I manage that? The part where I let it create the query for me, sounds too much magic :stuck_out_tongue:

Bakual:
> So I should keep with the helper approach because learning how to create a plugin takes longer than expected (I think).

Plugins aren't hard, but if you don't know them yet the helper should do as well.

> How do I manage that? The part where I let it create the query for me, sounds too much magic

Implement eg a "getListQuery" method into the helper which should return a JDatabaseQuery object. You could then expand on that query to add filter/sorting/limits if needed and run it. It obviously needs some standards there, eg the table name has to be aliased to a specified alias (eg "a") and the fields as well if they don't follow core standards (eg "name" is aliased as "title").
Or you can let the method directly return the items and pass filters or sorting as an argument.

jreys:
Hmmm, I think I'm getting the hang of it, but still, how do I handle extensions with multiple tables (with associations), inside the "getListQuery" having like a switch case for every table? even though with that approach I don't see how I can both perform queries and still have them on the multiple selection, sorry for all the questions but I just need to get this clear so I don't have to start over :)

Bakual:
The `getListQuery` method needs at least an argument which specifies the item type. Then the method has to adjust according to the requested type. In case of my own extension SermonSpeaker it would internally only change the table name and the rest of the query is similar.

To get the list of all possible item types, you need another method (eg `getItemTypes`) which returns an array of types.

There may of course be more to it. I haven't looked at every aspect of your proposal. That's just the listing part where I saw you did some quite complex stuff to get that data.

jreys:
Sorry for the late response, been busy lately, according to that will the queries made on the modelo be useless and be made on each component instead?

Bakual:
Yes, the component has to create the query. You may expand the query in com_associations (eg adding sorting and limiting statements) but that at least needs a standard in table name (alias). Eg the component should always use the "a" alias for the main table.

jreys:
So in practical terms, I pull the query from the component and add the filtering inside `../com_associations/models/associations.php` right? Also inside the `getListQuery()` should I add conditions for every item type the component supports? Categories I'm thinking of handling them inside the categories component since the query will always be the same (only the item type associated will change and that will be part of the query). If I'm getting something wrong please tell me :)

Bakual:
That sounds right to me.

infograf768:
@jreys 
Where do we stand on this now?

jreys:
Will work on it today and later with @alikon in person

infograf768:
??

administrator/components/com_contact/helpers/associations.php
```diff
@@ -0,0 +1,149 @@
...
+	 * Mapping of database columns alias.
+	 *
+	 * @return  array
+	 *
+	 * @since  __DEPLOY_VERSION__
+	 */
+	public static function getListQuery($itemType = '')
+	{
+		// Create a new query object.
+		$db = $this->getDbo();
```
infograf768: When using a public static method, I guess you should use `$db = JFactory::getDbo();` and not $this.
Same elsewhere.

infograf768:
Also, Select Item Type does not propose anything.
We should have I guess `if (file_exists($helpersPath . '/associations.php')`
and
`$columnAlias = $helper->getListQuery();`
But this last one will not work either as $this can't be used in `explode(', ', $this->getState(` as we are not in object context (we are in a helper and not a model).

infograf768:
@jreys 
This is still broken as I stated above. Can you solve it or do you need some help?

andrepereiradasilva:
i still don't see how this will work with components with more than 1 one item type with/without categories ...
again i will exemplificate a use case: A simple library component:
- Books (no associations)
- Books Categories (no associations)
- Authors (no associations)
- Authors Categories (no associations)
- Book Types **(with associations)**
- Book Types Categories **(with associations)**

Is not the component that supports or not associations is the item type itself ot it's category if it uses it

i mean the check needs to be done at item level

jreys:
Hmmmm, what about a new method? that returns an array of all the items and categories, and in the getListQuery() method, having the itemtype with a switch case for all items and if they have categories:

book_types, book_types.categories for example

